PR #17, opened 2021-07-30
Fix a problem with the schema for SNMPv1 Trap messages.

edarc:
Hello!

This fixes two issues with SNMPv1 `Trap` that were causing encode/decode errors, which were to do with the struct/schema definitions.

- `Trap` itself was missing a `#[rasn(choice)]` on `agent_addr`, resulting in an unconditional decoding error. 
- The schema for `IpAddress` was given as `[u8; 4]` which rasn derives as `SEQUENCE (SIZE (4)) OF INTEGER` rather than `OCTET STRING`; this is changed to `OctetString`.

This PR does not get this type 100% working, as there is at least one additional problem I've hit that I'm working to track down in another branch, which I think require changes to the derive macros. I believe it's to do with the handling of tags on tuple enum variants that themselves contain choice enums, but I'm still digging on it. There are notes and commented failing assertions in the added unit test indicating which situations trigger it.

### DIFF
--- a/standards/snmp/src/v1.rs
+++ b/standards/snmp/src/v1.rs
@@ -70,6 +70,7 @@ impl Pdu {
 #[rasn(tag(context, 4))]
 pub struct Trap {
     pub enterprise: ObjectIdentifier,
+    #[rasn(choice)]
     pub agent_addr: NetworkAddress,
     pub generic_trap: Integer,
     pub specific_trap: Integer,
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR! I'll review this in more depth soon.

standards/snmp/src/v1.rs
```diff
@@ -82,3 +83,111 @@ pub struct VarBind {
     pub name: ObjectName,
     pub value: ObjectSyntax,
```
XAMPPRocky, line 84:
```suggestion
    #[rasn(choice)]
    pub value: ObjectSyntax,
```

standards/snmp/src/v1.rs
```diff
@@ -82,3 +83,111 @@ pub struct VarBind {
...
+            .map(ToString::to_string)
+            .collect::<Vec<_>>()
+            .join(".")
+    }
+
+    #[test]
+    fn trap() {
+        #[cfg_attr(rustfmt, rustfmt_skip)]
+        let decode_data = [
+            0x30, 0x4f, // SEQUENCE -> Message
```
XAMPPRocky, line 106: Could this be formatted with the comments above the data rather than beside the ? I’m not a fan of putting comments on the same line.

standards/snmp/src/v1.rs
edarc: Done.

standards/snmp/src/v1.rs
edarc: Sure, done. I left the indentation for the nesting structure but I can also remove that if you prefer.

XAMPPRocky:
With those changes, the varbind encoding should work now, do you want to try re-enabling the checks?

edarc:
Thanks for the review!

This change does not entirely fix `VarBind` for me; I double-checked by trying to un-comment those assertions just to make sure.

The additional problems I'm hitting appear to be rooted in the derived impls of both `Encode` and `Decode` for choice enums that have one or more newtype variants where the contained type is itself a choice enum. For the case of `VarBind`, it appears that `ObjectSyntax` is the culprit as both of its variants are newtypes containing a choice enum.

I have managed to create a working fix for both derive macros that I'd like to send you in a separate PR once I get it cleaned up and more thoroughly tested, but to summarize what I _think_ I've worked out:

For `Encode`, the problem seems very simple -- it simply needs to generate calls to `encode` for newtype variants with an inner choice enum instead of `encode_with_tag`, similar to how it works for struct fields. I've fixed this by extending it to allow `#[rasn(choice)]` annotations on the variants to enable this behavior, analogous to choice struct fields.

`Decode` is a little trickier. Essentially, a newtype variant containing a choice enum can match an arbitrary number of tags, corresponding to the effective tags of each variant of the _contained_ choice enum. But the macro generated decode impl does not appear to handle this: choice enums derive `AsnType::TAG` as `EOC`, and newtype variants that have no explicit tag are dispatched on the tag of their contained type. Thus, a newtype variant containing a choice enum will be coded to dispatch on `EOC`, and thus no encoding of any variant of the contained type ever matches and decode always fails with "Invalid CHOICE discriminant".

The fix I've come up with for decoding is a little more involved, but essentially it splits the implementation of `decode` by partitioning the variants into choice and non-choice (meaning, newtype variants containing a choice enum annotated with `#[rasn(choice)]`, vs all other variants). My modified generated impl first attempts a tag dispatch as in the existing impl except there is an arm only for _non-choice_ variants. Then in the fall-through arm of the match, where currently it just bails with "Invalid CHOICE discriminant", it inserts a speculative attempt to decode as each choice variant, returning the first one that succeeds. Only then if all of the choice variants fail does the decode bail out.

That method of fixing it seemed the least invasive, but it has the downside that it does not fix the problem that such choice enums evade the static tag uniqueness check, because it does not recursively check the tags for choice enums contained in newtype variants. So if you have two newtype variants that contain choice enums, and the contained enums have variants with overlapping tags, I believe it will erroneously compile both before and after this change.

The explanation is a bit convoluted so let me know if I'm not describing it very well :) .. I'll try to get some branches up with the fixes I described above soon and you can decide if it's something you'd accept.

XAMPPRocky:
Thank you for the writeup, however I'm sure I agree that, that assessment of the derive macros is correct. The implementations for choice enums exist in the `Decode::decode`, and `Encode::encode` implementations, and the `EOC` tag is intentionally never actually used. You can see this yourself if you use `cargo-expand`.

edarc:
Indeed, `cargo-expand` helped me work out what was going on.

I pushed a new branch that has the fixes I came up with to make choice-enum-containing newtype variants encode and decode correctly as I described above. [Here is the diff](https://github.com/edarc/rasn/compare/fix-v1-trap...unnamed-choice-variants?expand=1).

Happy to send another PR for those changes if you think it's a reasonable way to fix the problem.

Thanks!

XAMPPRocky:
> Happy to send another PR for those changes if you think it's a reasonable way to fix the problem.

Well I think this is a reasonable approach, and I would accept it. However to be honest I was already kinda annoyed by having to mark it in the fields and variants `#[rasn(choice)]`. In my opinion it's annoying and I've missed it quite a few times, so I can imagine it would be easy for others to make this mistake, and the purpose of the library is to remove mistakes like this.

I haven't tested this yet, but if you'd be up for it, in this discussion I've come up with an idea that would allow us to remove the `#[rasn(choice)]` on fields and enums. The approach would be to add a associated constant to `AsnType` (`const CHOICE: bool = false`), that represents whether a type is a choice or not, and then in the derive implementation we instead change the codeine to generate the following for each variant/field. This would remove the need to tag it manually in the derive macro, and shouldn't cause any performance impact since the other branch will always be dead and can be removed by the compiler.

```rust
field = if <Field as rasn::AsnType>::CHOICE {
    Field::decode(...)
} else {
    Field::decode_with_tag(...)
};
```

XAMPPRocky:
Oh I just saw your edit, yeah nested choices are something that gives me a headache trying to resolve in the typesystem. 😄 That seems like the right approach for now until it is possible to recursively search a type's associated data.

edarc:
> Oh I just saw your edit, yeah nested choices are something that gives me a headache trying to resolve in the typesystem. 😄 That seems like the right approach for now until it is possible to recursively search a type's associated data.

Cool! Yes I tried for a little bit to figure out a way to recursively get the tags and never really got anywhere. (Perhaps it would need to wait for more stuff in Rust to be `const fn` in order to make it tractable.)

I agree, the `#[rasn(choice)]` attribute is annoying to have to remember. I think you're right that an associated constant might be able to make it go away so users don't have to remember to write it for their code to be correct.

Would you like me to send the existing solution using the `#[rasn(choice)]` annotations? I think it is largely independent from how the derive macro figures out which way to handle each field/variant, and I could try to retrofit the associated constant idea and get rid of the annoying annotation as a separate set of changes. I just don't want the PR to get too big and difficult for you to review.

XAMPPRocky:
> Would you like me to send the existing solution using the #[rasn(choice)] annotations? 

Sure, at the very least it resolves the problem for now.

XAMPPRocky:
Actually, give me the weekend, I think I might have figured it out.

edarc:
Closed as obsolete per discussion in #18.

PR #18, opened 2021-07-31
Fix encoding and decoding of newtype variants containing choice enums

edarc:
This contains the fixes we discussed on PR #17, **including** your idea for removing `#[rasn(choice)]` annotations on any newtype variant containing a choice enum. As you suggested, `AsnType` is instead extended to have an associated constant `const CHOICE: bool` defaulting to false. The macro for `AsnType` sets it to true in the derived impl for for choice enums.

Then, instead of chosing the handling of enum variants based on the attribute at macro-expand time, both choice and non-choice handling code is inserted into the macro expansion, controlled by conditionals referencing this const, and the compiler cleans out the dead branches in DCE.

In particular, for newtype variants, the tag match arm adds a pattern guard

```rust
TAG_n if !<ContainedType as rasn::AsnType>::CHOICE => {...}
```

which makes that arm dead code if `ContainedType` is a choice enum. Then, in the fall-through arm where no tag has matched, a speculative decode for each newtype variant is inserted, again guarded by a const-if that makes it dead code if the contained type is not a choice enum:

```rust
if <ContainedType as rasn::AsnType>::CHOICE {
    if let Ok(val) = ContainedType::decode(decoder) {
        return Ok(Ty::NewtypeVariant(val))
    }
}
```

And finally, for struct variants, each field's decoding contains a similar const conditional:

```rust
Ty::StructVariant {
    field: if <FieldType as rasn::AsnType>::CHOICE {
        <_>::decode(...)
    } else {
        <_>::decode_with_tag(...)
    },
}
```

Here's an [example `cargo expand` for SNMPv2 `VarBindValue`](https://gist.github.com/edarc/ef90168866456a2969b2998e9ef41080#file-expand-const-rs) showing what the new derived impl for `Decode` looks like. Derive of `Encode` uses the same idea but is much simpler.

I think this is pretty straightforward to extend to struct fields as well so the `#[rasn(choice)]` attribute can be eliminated for everything except top-level enum definitions. I haven't done that here but I am happy to work on it.

### DIFF
--- a/macros/src/asn_type.rs
+++ b/macros/src/asn_type.rs
@@ -93,6 +93,14 @@ pub fn derive_enum_impl(
 
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
 
+    let choice_const = if config.choice {
+        quote!(
+            const CHOICE: bool = true;
+        )
+    } else {
+        quote!()
+    };
+
     proc_macro2::TokenStream::from(quote! {
         #[automatically_derived]
         impl #impl_generics #crate_root::AsnType for #name #ty_generics #where_clause {
@@ -101,6 +109,7 @@ pub fn derive_enum_impl(
 
                 #tag
             };
+            #choice_const
         }
 
     })
--- a/macros/src/config.rs
+++ b/macros/src/config.rs
@@ -156,14 +156,12 @@ pub struct OptionalEnum {
 pub struct VariantConfig<'a> {
     variant: &'a syn::Variant,
     container_config: &'a Config,
-    pub choice: bool,
     pub tag: Option<Tag>,
 }
 
 impl<'a> VariantConfig<'a> {
     pub fn new(variant: &'a syn::Variant, container_config: &'a Config) -> Self {
         let mut tag = None;
-        let mut choice = false;
         let mut iter = variant
             .attrs
             .iter()
@@ -178,8 +176,6 @@ impl<'a> VariantConfig<'a> {
                 let path = item.path();
                 if path.is_ident("tag") {
                     tag = Tag::from_meta(item);
-                } else if path.is_ident("choice") {
-                    choice = true;
                 }
             }
         }
@@ -187,7 +183,6 @@ impl<'a> VariantConfig<'a> {
         Self {
             variant,
             container_config,
-            choice,
             tag,
         }
     }
--- a/macros/src/config.rs
+++ b/macros/src/config.rs
@@ -156,12 +156,14 @@ pub struct OptionalEnum {
 pub struct VariantConfig<'a> {
     variant: &'a syn::Variant,
     container_config: &'a Config,
+    pub choice: bool,
     pub tag: Option<Tag>,
 }
 
 impl<'a> VariantConfig<'a> {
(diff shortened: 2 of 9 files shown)


edarc:
Ahh sorry, I just saw your last update on #17. Feel free to close this if you've already got your own patchset for this.

XAMPPRocky:
Thank you for your PR! I've push up the changes onto `main`, do you want to try it out with #17 before we close this? Even if this doesn't patchiest doesn't make it in, I've added you as co-author to https://github.com/XAMPPRocky/rasn/commit/b2f0220d0f99b9187a3ad0aba0916b61236281d8 to credit this discussion for helping me finally figure it all out.

edarc:
> Thank you for your PR! I've push up the changes onto `main`, do you want to try it out with #17 before we close this?

Ooh, this TagTree solution is really interesting! I haven't grokked the whole thing yet but I think I get the idea how it works.

I locally rebased #17 onto the new changes and the unit test it adds for `Trap` is now failing to decode. I've tracked it down to the `agent_addr` field which is of type `NetworkAddress` (a choice enum), and the generated decode impl for it looks like this:

```rust
fn decode<D: rasn::Decoder>(decoder: &mut D) -> Result<Self, D::Error> {
    const TAG_0: rasn::Tag = rasn::Tag::new(rasn::types::Class::Context, 0usize as u32);
    if let Ok(value) = (|decoder: &mut D| {
        Ok::<_, D::Error>(NetworkAddress::Internet(<_>::decode_with_tag(
            decoder,
            rasn::Tag::new(rasn::types::Class::Context, 0usize as u32),
        )?))
    })(decoder)
    {
        return Ok(value);
    }
    Err(rasn::de::Error::custom(
        "Invalid `CHOICE` discriminant.",
    ))
}
```

`NetworkAddress` contains one newtype variant which has an `IpAddress` in it, which is tagged as application 0. But in the decode impl it's looking for context-specific 0 which causes it to fall through to `Err`. If I remove [this branch](https://github.com/XAMPPRocky/rasn/blob/b2f0220d0f99b9187a3ad0aba0916b61236281d8/macros/src/config.rs#L206-L207) it gives a correct impl for that type, but I'm not sure what that branch is there for originally, so I am not certain if removing it is actually the right fix.

Also, with that branch removed, it still cannot encode or decode `VarBind`, so the test only passes with same assertions skipped as in #17.

I'll do some more digging and see if any of the fixes from this PR are still relevant, or if it's now failing for a different reason entirely.

> I've added you as co-author to [b2f0220](https://github.com/XAMPPRocky/rasn/commit/b2f0220d0f99b9187a3ad0aba0916b61236281d8) to credit this discussion for helping me finally figure it all out.

Thanks, glad I could help! I appreciate your help with this, it's a really cool crate.

XAMPPRocky:
> Ooh, this TagTree solution is really interesting! I haven't grokked the whole thing yet but I think I get the idea how it works.

The latest main has more documentation and is more cleaned up. I'll work on fixing `VarBind`.

edarc:
Yes, all the issues I found seem to be resolved at main. Thanks! Closing this and #17.

PR #22, opened 2021-08-02
Small improvements to error messages, and clean up newly-dead tag constant defs

edarc:
Some small changes to add type information to the error messages about CHOICE enums, to help narrow down the search for issues with deeply-nested data structures.

Also, I noticed the `TAG_n` constant definitions emitted in `derive_enum_impl` are no longer referenced since the match block has been completely eliminated, so I removed them for a small cleanup.

### DIFF
--- a/macros/src/decode.rs
+++ b/macros/src/decode.rs
@@ -175,6 +175,10 @@ pub fn derive_enum_impl(
             }
         });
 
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR!

PR #23, opened 2021-08-03
Redo: Changed `slice_range(foo, _, foo.len())` to `slice_from(foo, _)`

rodrimati1992:
Disabled default feature of konst dependency, since no crate features are used.


XAMPPRocky:
Thank you for your PR!

PR #29, opened 2021-08-12
Fix OID encoding and decoding for BER and DER

### LINKED ISSUE #26: Invalid encoding and decoding of OIDs
Hello,
This crate is amazing and speeds up my small ASN.1 encoding project enormously :-)

However, I found a bug when encoding OIDs, having values bigger than 127. E.g. when taking following example ID from [MS website](https://docs.microsoft.com/en-us/windows/win32/seccertenroll/about-object-identifier) encoding using `rasn` gives back:
* **wrong** `[2b, 6, 1, 4, 1, b7, 2,15,14]` (this actually decodes to `1.3.6.1.4.1.7042.21.20`, look [here](http://lapo.it/asn1js/#MAsGCSsGAQQBtwIVFA))
* instead of **correct** `[2b, 6, 1, 4, 1, b7, 2, 15, 14]`

Another example is from [stackoverflow](https://stackoverflow.com/questions/5929050/how-does-asn-1-encode-an-object-identifier) and `1.2.840.113549.1.1.5 (sha1WithRsaEncryption)`:
* current (**incorrect**) encoding: `[2a, c8, 6, 8d, f7, 6, 1, 1, 5]` ([lapo.it](http://lapo.it/asn1js/#MAsGCSrIBo33BgEBBQ))
* **correct** encoding: `[2a, 86, 48, 86, f7, d, 1, 1, 5]` ([lapo.it](http://lapo.it/asn1js/#MAsGCSqGSIb3DQEBBQ))

BR, Piotr

zaczkows:
- `OID` value should be encoded using `base128`
- fixes #26


src/ber/enc.rs
```diff
@@ -5,7 +5,11 @@ use alloc::{borrow::ToOwned, collections::VecDeque, vec::Vec};
```
XAMPPRocky, line 5:
```suggestion
use alloc::vec;
```

XAMPPRocky:
It seems you also need to update this test with the correct encoding. https://github.com/XAMPPRocky/rasn/blob/main/standards/snmp/src/v1.rs#L152-L153

zaczkows:
Thank you for the info and suggestions. Now code and tests should be fixed.

XAMPPRocky:

PR #30, opened 2021-08-13
Making Implicit and Explicit value's public

### LINKED ISSUE #28: Implicit string value is private
When you have a Implicit string like below, it's not possible to access the value by itself because value is private.
`Implicit { _tag: PhantomData, value: "TestString" }`

```
error[E0616]: field `value` of struct `Implicit` is private
    |             println!("{:?}", obj.value);
    |                                  ^^^^^ private field

```

anthok:
Fixes  #28, allows access to both Implicit and Explicit value's.


XAMPPRocky:

PR #41, opened 2021-09-19
Add new example for RFC 7030 standard

zaczkows:
- add new item to standards: RFC7030 and reading of CSR attributes
- add missing `Eq`, `PartialOrd` and `Ord` for `Open` type; additionally add `ObjectIdentifier` to the enum
- allow to encode OIDs where second number is bigger than 39, e.g. 2.999.x is common OID in many examples
- update tests where appropriate
- somehow `cargo fmt` formatted code differently (newer version?)

### DIFF
--- a/standards/rfc7030/src/lib.rs
+++ b/standards/rfc7030/src/lib.rs
@@ -1,0 +1,206 @@
+/*
+From RFC7030 (https://datatracker.ietf.org/doc/html/rfc7030):
+CsrAttrs ::= SEQUENCE SIZE (0..MAX) OF AttrOrOID
+AttrOrOID ::= CHOICE { oid OBJECT IDENTIFIER, attribute Attribute }
+Attribute ::= SEQUENCE {
+        type  OBJECT IDENTIFIER,
+        values SET SIZE(0..MAX) OF ANY }
+*/
+
+#[cfg(test)]
+type CsrAttrs = Vec<AttrOrOid>;
+
+#[derive(rasn::AsnType, rasn::Decode, rasn::Encode, Debug, PartialEq, Clone)]
+#[rasn(choice)]
+enum AttrOrOid {
+    OID(rasn::types::ObjectIdentifier),
+    ATTRIBUTE(Attribute),
+}
+
+#[derive(rasn::AsnType, rasn::Decode, rasn::Encode, Debug, PartialEq, Clone)]
+struct Attribute {
+    r#type: rasn::types::ObjectIdentifier,
+    values: rasn::types::SetOf<rasn::types::Open>,
+}
+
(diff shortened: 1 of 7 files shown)


zaczkows:
There is still and _issue_ when encoding in DER format where items should be encoded based on their "encoded content length". Now, they depend on the order of the `Open` enum.

standards/rfc7030/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+#[rasn(choice)]
+enum AttrOrOid {
+    OID(rasn::types::ObjectIdentifier),
+    ATTRIBUTE(Attribute),
+}
+
+#[derive(rasn::AsnType, rasn::Decode, rasn::Encode, Debug, PartialEq, Clone)]
+struct Attribute {
+    r#type: rasn::types::ObjectIdentifier,
+    values: rasn::types::SetOf<rasn::types::Open>,
```
XAMPPRocky, line 23: This should be using `Any` not `Open`. `Open` can't handle things like sequences, where as `Any` can.

standards/rfc7030/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+*/
+
+#[cfg(test)]
+type CsrAttrs = Vec<AttrOrOid>;
+
+#[derive(rasn::AsnType, rasn::Decode, rasn::Encode, Debug, PartialEq, Clone)]
+#[rasn(choice)]
+enum AttrOrOid {
+    OID(rasn::types::ObjectIdentifier),
+    ATTRIBUTE(Attribute),
```
XAMPPRocky, line 17: This should be in pascal case like it is in the PKIX crate.

XAMPPRocky (commented):
Thank you for your PR! I have a couple of code comments, and I also was wondering with the PKIX crate I already have, would this be more useful as a submodule in that? What do you think?

standards/rfc7030/src/lib.rs
zaczkows: done

XAMPPRocky:
Sorry my suggestion might not have been clear. When I said sub module I meant more like what if this would be declared as `mod rfc7030;` or `mod est;` inside the `rasn-pkix`, rather than its own crate?

XAMPPRocky:
Also there's a couple of items missing from the module defined in the errata.

https://datatracker.ietf.org/doc/html/rfc8951#appendix-A

zaczkows:
I hope it looks better now :-)

XAMPPRocky:
Thank you for your PR! There's a couple of small things I'd to change, but this is good enough to merge, I can update those myself.

PR #47, opened 2021-09-29
Added CMS standard: https://datatracker.ietf.org/doc/html/rfc5652

ancwrd1:
This PR adds CMS (a.k.a. PKCS#7) data types used with S/MIME messaging.


standards/cms/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+pub const OID_SIGNING_TIME: ConstOid = ConstOid(&[1, 2, 840, 113549, 1, 9, 5]);
+pub const OID_COUNTER_SIGNATURE: ConstOid = ConstOid(&[1, 2, 840, 113549, 1, 9, 6]);
+
+// content types
+pub const OID_CONTENT_DATA: ConstOid = ConstOid(&[1, 2, 840, 113549, 1, 7, 1]);
+pub const OID_CONTENT_SIGNED_DATA: ConstOid = ConstOid(&[1, 2, 840, 113549, 1, 7, 2]);
+pub const OID_CONTENT_ENVELOPED_DATA: ConstOid = ConstOid(&[1, 2, 840, 113549, 1, 7, 3]);
+pub const OID_CONTENT_DIGESTED_DATA: ConstOid = ConstOid(&[1, 2, 840, 113549, 1, 7, 5]);
+pub const OID_CONTENT_ENCRYPTED_DATA: ConstOid = ConstOid(&[1, 2, 840, 113549, 1, 7, 6]);
+pub const OID_CONTENT_AUTHENTICATED_DATA: ConstOid = ConstOid(&[1, 2, 840, 113549, 1, 9, 16, 1, 2]);
```
XAMPPRocky, line 29: Can we add these OIDs as consts in the main crate, and re-export them here instead similar to the other crates?

XAMPPRocky (commented):
Thank you for your PR! This mostly looks good to me I have a couple of things.

- Can we add some documentation to the types instead of comments showing the ASN.1 syntax? I think that would be more useful.
- What would you think about having this as `pub mod cms;` module in the PKIX crate similar to EST?

standards/cms/src/lib.rs
ancwrd1: > Can we add some documentation to the types instead of comments showing the ASN.1 syntax? I think that would be more useful.

Yeah the current comments there are not rustdocs exactly, they are mainly added to visually check whether the structs match the ASN.1 syntax because it's rather hard to cover it fully with unit tests considering the CMS complexity. I'll go ahead and do some copy/paste work from the RFC.

> What would you think about having this as pub mod cms; module in the PKIX crate similar to EST?

I'd argue that CMS is not directly a sub-standard of PKIX but rather depends on it, because for example it may be used without certificates at all.

> Can we add these OIDs as consts in the main crate, and re-export them here instead similar to the other crates?

Sure, will do. Do you want a compliant naming as well, e.g. `ISO_MEMBER_BODY_US_RSADSI_xxx`?

ancwrd1: Fixes added as requested.

XAMPPRocky:

XAMPPRocky:
I've now published this on crates.io https://docs.rs/rasn-cms

Issue #61, opened 2022-01-22
microsoft seems to use a different EncapsulatedContentInfo struct


dvc94ch:
microsoft msix and appx packages contain a `CodeIntegrity.cat` and `AppxSignature.p7x` file, which while undocumented seems to be cms encoded SignedData. however it seems like it's not quite spec compliant and maybe has an extra field or something.

```
thread 'test_msix_p7x' panicked at 'called `Result::unwrap()` on an `Err` value: FieldError { name: "SignedData.encap_content_info", error: "Field `EncapsulatedContentInfo.content`: Unexpected extra data found: length `264` bytes" }', standards/cms/tests/test_cms.rs:44:62
```
```
thread 'test_msix_cat' panicked at 'called `Result::unwrap()` on an `Err` value: FieldError { name: "SignedData.encap_content_info", error: "Field `EncapsulatedContentInfo.content`: Unexpected extra data found: length `1303` bytes" }', standards/cms/tests/test_cms.rs:58:62
```

replacing content with Any works.

dvc94ch:
This seems to fix the issue without breaking the existing tests. @XAMPPRocky is this a sensible change?
```diff
diff --git a/standards/cms/src/lib.rs b/standards/cms/src/lib.rs
index 55d006f..14bddab 100644
--- a/standards/cms/src/lib.rs
+++ b/standards/cms/src/lib.rs
@@ -208,7 +208,7 @@ pub struct OtherRevocationInfoFormat {
 #[derive(AsnType, Clone, Debug, Decode, Encode, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct EncapsulatedContentInfo {
     pub content_type: ContentType,
-    #[rasn(tag(explicit(0)))]
+    #[rasn(tag(context, 0))]
     pub content: Option<OctetString>,
 }
```

XAMPPRocky:
Thank you for your issue! Right now I don't think I would want to make the change, since the IETF RFC is what should be followed, rather than Microsoft's undocumented formats. I would prefer to add a seperate type for Microsoft types over having spec non-compliant types.

dvc94ch:
from RFC5652 it is aknowledged that cms breaks compatibility with pkcs7, which is also an IETF RFC.

```
5.2.1.  Compatibility with PKCS #7

   This section contains a word of warning to implementers that wish to
   support both the CMS and PKCS #7 [PKCS#7] SignedData content types.
[...]

PR #62, opened 2022-01-22
Add failing tests.

dvc94ch:
tests for #61

### DIFF
--- a/standards/cms/src/pkcs7_compat.rs
+++ b/standards/cms/src/pkcs7_compat.rs
@@ -1,0 +1,61 @@
+//! PKCS7 compatibility module.
+//!
+//! The Cryptographic Message Syntax and PKCS7 are mostly the same. The only
+//! difference is that the `EncapsulatedContentInfo` content field has the `Any`
+//! type instead of the `OctetString`. See section 5.2.1. Compatibility with PKCS #7
+//! of RFC5652 for further information.
+use crate::{
+    Any, AuthAttributes, CertificateSet, CmsVersion, ContentType, Digest,
+    DigestAlgorithmIdentifier, DigestAlgorithmIdentifiers, MessageAuthenticationCode,
+    MessageAuthenticationCodeAlgorithm, OriginatorInfo, RecipientInfos, RevocationInfoChoices,
+    SignerInfos, UnauthAttributes,
+};
+use rasn::{AsnType, Decode, Encode};
+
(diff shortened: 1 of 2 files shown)


dvc94ch:
so looks like pkcs7 ~= cms but with the EncapsuledContentInfo.content being `Any` instead of an `OctetString`. Is changing it to `Any` for compatibility ok?

XAMPPRocky:
Could you provide links showing the two differences? I'm hesitant to change the type since this is based on the ietf RFC.

dvc94ch:
removed the tests and added a `pkcs7_compat` module as discussed in #61

standards/cms/src/pkcs7_compat.rs
```diff
@@ -0,0 +1,54 @@
+use crate::{
```
XAMPPRocky, line 1: Could we add some top level documentation about this module? Just a short description about the incompatibility

standards/cms/src/pkcs7_compat.rs
```diff
@@ -0,0 +1,60 @@
...
+    MessageAuthenticationCodeAlgorithm, OriginatorInfo, RecipientInfos, RevocationInfoChoices,
+    SignerInfos, UnauthAttributes,
+};
+use rasn::{AsnType, Decode, Encode};
+
+/// The content is represented in the type EncapsulatedContentInfo
+#[derive(AsnType, Clone, Debug, Decode, Encode, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct EncapsulatedContentInfo {
+    pub content_type: ContentType,
+    pub content: Any,
```
XAMPPRocky, line 19: Shouldn't this have the following?
```suggestion
    #[rasn(tag(explicit(0))]
    pub content: Any,
```

dvc94ch: Still seems to work with this change, what does this do?

XAMPPRocky: Well it will not be parsed correctly otherwise, as the `Any` type will now contain `EXPLICIT [0]` sequence, so when you try to decode the `Any` type as some other type, it will then fail.

dvc94ch: looks like that's right according to the RFC I linked, thanks!

XAMPPRocky:

PR #75, opened 2022-03-02
updating test to full as_req

x-delfino:
I've updated the as_req test to include a valid as_req


XAMPPRocky:

PR #86, opened 2022-03-17
Convert `Result` in macros to `core::result::Result`

orthecreedence:
The idea here is when using the #[derive()] macros, we don't interfere with libraries that have aliased `Result` to `Result<T, MyError>`.


XAMPPRocky:

PR #92, opened 2022-04-12
wip: very very basic compiler implementation

masihyeganeh:
It can convert this `asn1` schema:
```asn1
Simple DEFINITIONS ::=
BEGIN
  SimpleSequence ::= SEQUENCE {
    someInt INTEGER,
    someBytes OCTET STRING
  }

  MoreComplexSequence ::= SEQUENCE {
    someOptionalInt INTEGER OPTIONAL,
    someIntEnum INTEGER {value1(123), value2(456)},
    someOptionalBytes OCTET STRING OPTIONAL,
    someOptionalReference SimpleSequence OPTIONAL,
    someOptionalReferenceWithContextNumber [2] SimpleSequence OPTIONAL,
    someAnonymousSequence SEQUENCE {
        simpleInt INTEGER
    }
  }

  MainSequence ::= [APPLICATION 1] EXPLICIT SEQUENCE {
    simpleSequence SimpleSequence,
    moreComplexSequence MoreComplexSequence
  }
END
```

To this `Rust` code:
```rust
use rasn::types::OctetString;
use rasn::types::Integer;
use rasn::Encode;
use rasn::Decode;
use rasn::AsnType;
#[non_exhaustive]
pub struct SomeIntEnum;
impl SomeIntEnum {
	pub Value2: Integer = 456;
	pub Value1: Integer = 123;
}

#[derive(AsnType, Encode, Decode, PartialEq, PartialOrd, Eq, Ord, Debug)]
#[rasn(tag(explicit(application, 1)))]
pub struct MainSequence {
	pub simple_sequence: SimpleSequence,
	pub more_complex_sequence: MoreComplexSequence,
}

#[derive(AsnType, Encode, Decode, PartialEq, PartialOrd, Eq, Ord, Debug)]
pub struct SomeAnonymousSequence {
	pub simple_int: Integer,
}

#[derive(AsnType, Encode, Decode, PartialEq, PartialOrd, Eq, Ord, Debug)]
pub struct MoreComplexSequence {
	pub some_optional_int: Option<Integer>,
	pub some_int_enum: Integer,
	pub some_optional_bytes: Option<OctetString>,
	pub some_optional_reference: Option<SimpleSequence>,
	#[rasn(tag(2))]
	pub some_optional_reference_with_context_number: Option<SimpleSequence>,
	pub some_anonymous_sequence: SomeAnonymousSequence,
}

#[derive(AsnType, Encode, Decode, PartialEq, PartialOrd, Eq, Ord, Debug)]
pub struct SimpleSequence {
	pub some_int: Integer,
	pub some_bytes: OctetString,
}
```

### DIFF
--- a/compiler/src/codegen.rs
+++ b/compiler/src/codegen.rs
@@ -1,4 +1,5 @@
 mod constant;
+mod constant_enum;
 mod imports;
 mod structs;
 
@@ -7,11 +8,13 @@ use std::{collections::HashSet, fmt, io::Write, mem};
 use failure::Fallible as Result;
 use heck::*;
 
-use self::{constant::Constant, imports::*, structs::*};
-use crate::{
-    parser::*,
-    semantics::SemanticChecker,
+use self::{
+    constant::Constant,
+    imports::*,
+    structs::{Field as StructField, *},
 };
+use crate::codegen::constant_enum::{ConstantEnum, Triple};
+use crate::{parser::*, semantics::SemanticChecker};
 
 #[derive(Clone, Copy, Debug)]
 pub enum TagEnvironment {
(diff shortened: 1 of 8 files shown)


compiler/src/codegen.rs
```diff
@@ -1,4 +1,5 @@
 mod constant;
+mod constant_enum;
```
masihyeganeh, line 2: The whole `constant_enum` is just another version of `constant` but I think the output is easier to use.
For Example for `version INTEGER {v3(3)}(v3,...),` it produces a `Version` struct with `V3` field, So you can easily write `Version::V3` wherever you need the `3`.

compiler/src/codegen.rs
```diff
@@ -40,19 +43,50 @@ impl Default for TagEnvironment {
 
 pub trait Backend: Default {
     fn tag_environment(&mut self, environment: TagEnvironment);
-    fn generate_type(&mut self, ty: &Type) -> Result<String>;
+    fn generate_type(
+        &mut self,
+        ty: &Type,
+        parent_prefix: Option<&Prefix>,
```
masihyeganeh, line 49: I HATE this, but for prefixes of sequences, like `MainSequence ::= [APPLICATION 1] EXPLICIT SEQUENCE {...}`, the `APPLICATION 1` is parsed by the caller, but it is needed here, so it is passed down with each function call.

compiler/src/codegen.rs
```diff
@@ -40,19 +43,50 @@ impl Default for TagEnvironment {
...
+    ) -> Result<(String, Option<Prefix>)>;
```
masihyeganeh, line 50: I HATE this too, but for prefixed of fields, like `someFieldWithContextNumber [2] INTEGER`, the `[2] INTEGER` is parsed here but the caller has already written the field name and we can't add a macro call here, so we need to return extra data to caller, so it can write the macro call in the right place.

compiler/src/codegen/structs.rs
```diff
@@ -138,17 +178,70 @@ pub enum Serde {
...
+            Derive::Ord => "Ord",
+            Derive::Debug => "Debug",
         };
 
         derive.fmt(f)
     }
 }
+
+pub enum Rasn {
+    Type(&'static str),
```
masihyeganeh, line 209: Maybe we can use `BuiltinType` here instead of `&str`. Or not

XAMPPRocky:
Thank you for your PR, and congrats on your first contribution! 🎉 I'm not too worried about the details, as I think we'll change them a lot as this is worked on and modernised.

PR #102, opened 2022-05-12
Added Microsoft Authenticode structures

ancwrd1:
This PR adds Microsoft Authenticode structures based on PKCS7. They are used for signing PE executables.


XAMPPRocky:
Thank you for your PR!

PR #106, opened 2022-07-08
Allow buffer reuse in `Encoder`

### LINKED ISSUE #72: Allow using user-supplied buffer for encoding data?
Hi, I'm working on building a SNMP client and as an optimization to prevent allocating memory on every send/receive, I want to internally store a `Vec` in my client, however there doesn't appear to be a way to provide the `Encoder` a `&mut Vec<u8>` or `&mut [u8]` to encode into, which would be really useful!

repnop:
Fixes #72 

I think this should be sufficient for my use case, however I did have a question. I originally had a `new_set_with_buffer` but as `self.output()` doesn't actually use the `self.output` for the final encoding when `is_set_encoding: true` and is `mem::take(&mut self.output)`'d elsewhere, I felt like it would be pretty useless since the whole point is to allow reusing the buffer space for the final encoding, but this does kind of expose this implementation detail and could cause some confusion since there's a missing function, not entirely sure what a good solution is for this.

Let me know if there's anything else you'd like adjusted.

### DIFF
--- a/src/ber/enc.rs
+++ b/src/ber/enc.rs
@@ -57,6 +57,15 @@ impl Encoder {
         }
     }
 
+    /// Creates a new instance from the given `config` and a user-supplied
+    /// `Vec<u8>` buffer. This allows reuse of an existing buffer instead of
+    /// allocating a new encoding buffer each time an [`Encoder`] is created.
+    /// The buffer will be cleared before use.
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

XAMPPRocky:
Also congrats on finding an ICE. 🙂

repnop:
well now, I didn't think my PR would do that much damage :smile:

PR #118, opened 2023-04-30
Look at data string and pick best guess parse format. Does not yet ha…

jsatchell:
…ndle case where optional seconds or minutes are absent.

### DIFF
--- a/src/ber/de.rs
+++ b/src/ber/de.rs
@@ -313,36 +313,57 @@ impl<'input> crate::Decoder for Decoder<'input> {
             if string.contains('.') {
                 "%Y%m%d%H%M%S%.3fZ"
             } else {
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR! Sorry for the delay, I've disabled the mips targets so if you rebase it should pass and then be good to merge.

jsatchell:
Now done the optional cases, where seconds or minutes are absent. I think the code now handles all the variants.

XAMPPRocky:
Damn android, if you rebase again it should be good to go 🙂

jsatchell:
Looks hopeful. In ideal world, we should have unit test coverage for all variants of everything.

XAMPPRocky:
Thank you for your PR!

PR #121, opened 2023-06-26
Fix tag on PKIX CRL extension

zkonge:
Without `explicit`, it returns empty vec for all extensions, and can't parse any existed CRL properly.

e.g. http://crls.pki.goog/gts1c3/zdATt0Ex_Fk.crl


XAMPPRocky:

PR #130, opened 2023-08-01
fix: remove duplicates in PrintableString char set

6d7a:
Great library!

Removed duplicates of 'E' and 'e' characters in the `PrintableString` character set


XAMPPRocky:

PR #132, opened 2023-08-01
Fix Clippy warnings

Nicceboy:
Fix Clippy warnings on default configuration.

There seems to be also quite many warnings on pedantic level when checking with `﻿﻿cargo clippy -- -W clippy::pedantic` 

Would it be okay to also attempt to fix some of these?


XAMPPRocky:

PR #135, opened 2023-08-02
correctly handle non-byte-aligned BIT STRING length

### LINKED ISSUE #134: DER encoded `BIT STRING` produces `BitString` with incorrect length
Decoding a DER-encoded `BIT STRING` with a non-zero "unused bits" byte into a `types::BitString` results in a value padded with `0`s to the nearest byte.

This makes it impossible to tell the difference between values with "real" trailing zeros and those without.

In particular, [RFC3779] and other RPKI-related standards use a `BIT STRING` to encode IP prefixes, indicating the length of the prefix through the bit-length of the `BIT STRING`. This issue makes it impossible to correctly decode IP prefixes with lengths that are not a multiple of 8.

Note that I have *only* tried this with the DER implementation. I do not know whether the other codecs suffer from a similar issue.

The issue can be reproduced with the following minimal example:

``` rust
use bitvec::{bitvec, order::Msb0};
use rasn::{der, types::BitString};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // DER-encoded byte-slice of BIT STRING B'0101'
    let encoded: &[u8] = &[
        0x03,        // tag = 3
        0x02,        // length = 2
        0x04,        // unused bits = 4
        0b0101_0000, // value
    ];
    let bit_string = der::decode::<BitString>(encoded)?;

    // Expected value
    let expected = bitvec![u8, Msb0; 0, 1, 0, 1];

    assert_eq!(bit_string, expected);

    Ok(())
}
```

I will take a look at the relevant code, and propose a fix if I can grok it easily enough.

[RFC3779]: https://datatracker.ietf.org/doc/html/rfc3779#section-2.2.3.8

benmaddison:
Fixes #134

### DIFF
--- a/src/ber/de.rs
+++ b/src/ber/de.rs
@@ -268,12 +268,10 @@ impl<'input> crate::Decoder for Decoder<'input> {
                         }
 
                         let mut string = types::BitString::from_vec(buffer);
-                        // TODO:
-                        // Using `saturating_sub` here ensures that the `flip1` test passes,
-                        // but returning an error on an overflow is probably more correct,
-                        // at least for DER.
-                        // Need to consult the spec...
-                        let bit_length = string.len().saturating_sub(bits as usize);
+                        let bit_length = string
+                            .len()
+                            .checked_sub(bits as usize)
+                            .ok_or(Error::InvalidBitString { bits })?;
                         string.truncate(bit_length);
 
                         Ok(string)
--- a/src/ber/de.rs
+++ b/src/ber/de.rs
@@ -267,13 +267,16 @@ impl<'input> crate::Decoder for Decoder<'input> {
                             *last &= !((1 << bits) - 1);
                         }
 
-                        let string = types::BitString::from_vec(buffer);
+                        let mut string = types::BitString::from_vec(buffer);
+                        // TODO:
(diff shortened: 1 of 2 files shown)


src/ber/de.rs
```diff
@@ -267,13 +267,16 @@ impl<'input> crate::Decoder for Decoder<'input> {
...
-                            Ok(string)
-                        }
+                        let mut string = types::BitString::from_vec(buffer);
+                        // TODO:
+                        // Using `saturating_sub` here ensures that the `flip1` test passes,
+                        // but returning an error on an overflow is probably more correct,
+                        // at least for DER.
+                        // Need to consult the spec...
+                        let bit_length = string.len().saturating_sub(bits as usize);
+                        string.truncate(bit_length);
```
XAMPPRocky, line 277: I'd switch this to `checked_sub` and then return an error, I'm pretty sure the spec doesn't cover what you should do here

benmaddison: I've taken a look at X.690 again:
The only way that we can get an overflow here is if `string.len() == 0 && bits > 0`, meaning the BIT STRING was empty but the unused bits octet was non-zero.
Thus clause 8.6.2.3 applies, and the right thing to do is error.
Will do this now...

XAMPPRocky:

PR #137, opened 2023-08-07
Draft: Possibly some safer arithmetics

Nicceboy:
Hey,

I ran the Facebook's MIRAI  static analysis tool for this project, and it identified couple issues which might be real or not.

I added boundary checks for some arithmetics as a result, which looked like potential problem scenarios. I do not understand to code well enough to say for sure. Or should we create tests for these?

Adding overflow check might come with with small performance penalties. So it might be good to look if some are indeed relevant.


XAMPPRocky:
Thank you for your PR!

PR #140, opened 2023-08-08
Fix OID validation, add docs build for CI, prevent spaces and globbin…

Nicceboy:
There seems to be OIDs with array size of 1 while the valid OID is checked to require size of 2 at least.
This has prevented docs from building.

I modified the requirement to just match that array is not empty.
Is this correct way?

I also added test for CI to check that docs will build before they end up to crates.io


XAMPPRocky:
Thank you for your PR!

PR #142, opened 2023-08-15
Improved size constraint

Nicceboy:
A small suggestion for size constraint to make implementation of codes easier.

Makes underlying Bounded object public and adds few functions to detect the type of the Size.

I also added `#[must_use]` attributes, since calling these without using the result does not make sense.


XAMPPRocky:
Thank you for your PR!

PR #144, opened 2023-08-17
Split OID codec functionality in BER to be usable by other codecs

Nicceboy:
Other codecs seems to re-use bytepresentation of BER encoding for ObjectIdentifier type. 

This change would make it easier to access bytes without added length or tags.

### DIFF
--- a/src/ber.rs
+++ b/src/ber.rs
@@ -9,15 +9,32 @@ pub use identifier::Identifier;
 pub(crate) use rules::EncodingRules;
 
 /// Attempts to decode `T` from `input` using BER.
+/// # Errors
+/// Returns error specific to BER decoder if decoding is not possible.
 pub fn decode<T: crate::Decode>(input: &[u8]) -> Result<T, de::Error> {
     T::decode(&mut de::Decoder::new(input, de::DecoderOptions::ber()))
 }
 
 /// Attempts to encode `value` to BER.
+/// # Errors
+/// Returns error specific to BER encoder if encoding is not possible.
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
> This change would make it easier to access bytes without added length or tags.

Thank you for your PR! Are you sure the length and tag should not be included other codecs? In the specs it's often cited as being a complete representation of the BER value, which would include tag and length.

Nicceboy:
The existing functions are still there which can be used to get the tag and length as well, if required. I do not know all the codecs. I noticed that in UPER you used the tag and length at least. 

I compared to other implementations of OER and at least in this case they seems to drop. 
OER adds its own length determinant.

In standard it says that "***contents octets*** of BER encoding of the object identifier value" (specified in 8.19), so I would assume that is is without length and tags. 

In 8.19.5 section there is an example, which distincts contents

OBJECT IDENTIFIER | Length | Contents |
-|-|-|
06<sub>16</sub>| 03<sub>16</sub> | 883703<sub>16</sub>

Nicceboy:
Seems like that similar wording is also used in PER. Apparently the logic should be the same but which one it is 😄

Nicceboy:
I think this also applies for types `GeneralizedTime` and `UTCTime`. Should we split these functions in BER as well, or do helper functions outside of BER? Parsing the time strings would be identical.

XAMPPRocky:
>  Should we split these functions in BER as well, or do helper functions outside of BER? Parsing the time strings would be identical.

In BER so it's clear that it's the BER encoding being used.

XAMPPRocky:
Thank you for your PR!

PR #149, opened 2023-08-24
Add TryFrom for str type for PrintableString

Nicceboy:
Also small documentation updates, enforce use of `from` or `try_from`, same trait order

### DIFF
--- a/src/types/strings/printable.rs
+++ b/src/types/strings/printable.rs
@@ -2,11 +2,16 @@ use super::*;
 
 use alloc::{borrow::ToOwned, boxed::Box, string::String, vec::Vec};
 
-/// A string, which contains the characters defined in "X.680 Table 10".
+/// A string, which contains the characters defined in X.680 41.4 Section, Table 10.
+///
+/// You must use `try_from` or `from_*` to construct a `PrintableString`.
 #[derive(Debug, Default, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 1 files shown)


src/types/strings/printable.rs
```diff
@@ -15,21 +20,25 @@ impl StaticPermittedAlphabet for PrintableString {
...
-            "{} not in character set",
-            ch
+            "{ch} not in character set"
         );
         self.0.push(ch as u8);
     }
+
+    fn chars(&self) -> Box<dyn Iterator<Item = u32> + '_> {
+        Box::from(self.0.iter().map(|byte| *byte as u32))
+    }
```
XAMPPRocky, line 33: Why `u32` when it can fit in a `u8`? Also why use `Box<dyn Iterator>` instead of `impl Iterator`?

Nicceboy: Those were already there, I just changed the order to be same as in Trait. I have no idea 😅 

https://github.com/XAMPPRocky/rasn/blob/a7a00c76d702e530a14d517d830adc899bd31dc4/src/types/strings/printable.rs#L18-L20

I can change those at the same time.

XAMPPRocky: Oh I didn't notice this was in the trait, in that case it makes sense because you can't use either of those.

XAMPPRocky:
Thank you for your PR!

PR #150, opened 2023-08-25
Improves BER GeneralizedTime decoders, support for CER/DER strict mode

### LINKED ISSUE #57: `GeneralizedTime` encodes unaccepted value.
My fuzzer for this was

```rust
#![no_main]
use libfuzzer_sys::fuzz_target;

fuzz_target!(|data: &[u8]| {
    if let Ok(value) = rasn::der::decode::<rasn::types::Open>(data) {
        assert_eq!(value, rasn::der::decode(&rasn::der::encode(&value).unwrap()).unwrap());
    }
});
```

Not sure if this a valid round fuzz test, feel free to close as invalid if not.

Reproducer is

```rust
fn main() {
    let data = [
        24, 19, 43, 53, 49, 54, 49, 53, 32, 32, 48, 53, 50, 52, 48, 57, 52, 48, 50, 48, 90,
    ];

    let value = rasn::der::decode::<rasn::types::Open>(&data).expect("passes");

    rasn::der::decode::<rasn::types::Open>(&rasn::der::encode(&value).unwrap()).expect("fails");
}
```

Fails with `thread 'main' panicked at 'fails: NoValidChoice { name: "Open" }', src/main.rs:7:81`

Nicceboy:
BER GeneralizedTime decoder should now handle all cases.
Added stricter function also for CER/DER. 
I am not sure, if we should use BER decoder for CER/DER as well, but at least asn1tools use stricter versions. 

Resolves #57 

Also Encoder produces valid canonical form now.

Make these functions public, so that other codecs can access them.

Since BER allows any ASN.1 compatible GeneralizedTime type, should this check to be moved into the type itself?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ repository = "https://github.com/XAMPPRocky/rasn.git"
 
 [workspace.dependencies]
 pretty_assertions = "1.1"
-chrono = { version = "0.4.19", default-features = false, features = ["alloc"] }
+chrono = { version = "0.4.26", default-features = false, features = ["alloc"] }
 bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 
 [features]
--- a/src/ber.rs
+++ b/src/ber.rs
@@ -196,4 +196,91 @@ mod tests {
             }
         }
     }
+    #[test]
+    fn test_generalized_time() {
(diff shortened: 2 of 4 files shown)


XAMPPRocky:
Thank you for your PR! 

> Since BER allows any ASN.1 compatible GeneralizedTime type, should this check to be moved into the type itself?

Yes, let's make that a follow-up PR though.

PR #152, opened 2023-08-26
ber: Split and improve UtcTime encode/decode

Nicceboy:
Splits functionality similarly than for GeneralizedTime.

Restricted decode for CER/DER.

There was a bug when decoding time offset values, `991231235959+0000` is corresponding value for `991231235959Z` but in tests it was `991231235959+0200`.


XAMPPRocky:
Thank you for your PR!

PR #156, opened 2023-09-06
Fix/constrained extension

6d7a:
I ran into a problem with constrained extension additions the other day. I represented the following ASN1 snippet...
```asn1
TestSequence ::= SEQUENCE { 
     hello OCTET STRING (SIZE(0..8)),
     ...,
     world INTEGER(0..9) DEFAULT 8
}
```
...using rasn as...
```rust
#[derive(AsnType, Debug, Clone, Decode, Encode, PartialEq)]
#[non_exhaustive]
struct TestSequence {
      #[rasn(size("0..=8"))]
      hello: OctetString,
      #[rasn(
            extension_addition,
            value("0..=9"),
            default = "test_sequence_world_default"
      )]
      world: u8,
}

fn test_sequence_world_default() -> u8 {
     8
}
```
...and discovered that an encoding-decoding round trip fails. I did some digging and it turned out that the decoder does not take constraints into account when dealing with extension additions. This PR provides the test case above and should fix the issue. The crucial fix is in the extension addition function of the per-decoder in `src/per/de.rs`, where I am passing along constraint information.

### DIFF
--- a/macros/src/config.rs
+++ b/macros/src/config.rs
@@ -1011,8 +1011,40 @@ impl<'a> FieldConfig<'a> {
         };
 
         if self.extension_addition {
-            quote! {
-                decoder.decode_extension_addition() #or_else #handle_extension
+            match (
+                self.default.as_ref().map(|path| {
+                    path.as_ref()
+                        .map_or(quote!(<_>::default), |path| quote!(#path))
+                }),
+                self.constraints.has_constraints(),
+            ) {
+                (Some(path), true) => {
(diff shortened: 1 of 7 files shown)


6d7a:
I also added the `encode_default_with_constraints` that is invoked in `macros/src/config.rs:876`.

XAMPPRocky:
Thank you for your PR and congrats on your first contribution! I thought I had done this but I realised I mostly focused on constraints through the type system.

PR #158, opened 2023-09-22
Fix/issue 141

### LINKED ISSUE #141: Unable to decode Option<Time>, or generally, Option<enum>
# problem
related: https://github.com/XAMPPRocky/rasn/commit/63ac735d78b0b51e5ee0320d597650edbb2dc49d

The commit was patched on 0.7, and there is a bit change at https://github.com/XAMPPRocky/rasn/commit/8068f77e0d70f70f17dc647e53506c34e7d185b6

I only test it with code from crates.io version and not fully tested it with GitHub master branch. So problem still exists (but is a new problem).

It caused in https://github.com/XAMPPRocky/rasn/blob/e64391c49c34efdd3a342df81976a56cdfb487d2/standards/pkix/src/lib.rs#L359

reproduced by: 

```rust
    let s = hex::decode("170d3730303130313030303031305a").unwrap();
    let t: Option<Time> = rasn::der::decode(s.as_slice()).unwrap();
    dbg!(t);
    // None
    let t: Time = rasn::der::decode(s.as_slice()).unwrap();
    dbg!(t);
    // Utc(1970-01-01T00:00:10Z)
```
# reason

I believe the reason happened here: 
https://github.com/XAMPPRocky/rasn/blob/e64391c49c34efdd3a342df81976a56cdfb487d2/src/ber/de.rs#L76-L85

It doesn't check if the input is a CHOICE, for all choices, it failed.

Quick fix, I'm not familiar with the other part of the lib and think **it's wrong** (streaming use it too?), but it make the code above run:
```rust
fn parse_constructed_contents<D, F>(
    &mut self,
    tag: Tag,
    check_identifier: bool,
    decode_fn: F,
) -> Result<D>
    where
        F: FnOnce(&mut Self) -> Result<D>,
{
    if let Tag::EOC = tag {
        return (decode_fn)(self);
    }
    let (identifier, contents) = self.parse_value(tag)?;
```
# by the way
there is a typo :P
https://github.com/XAMPPRocky/rasn/blob/e64391c49c34efdd3a342df81976a56cdfb487d2/standards/pkix/src/lib.rs#L342

6d7a:
Closes #141 

There was indeed a check for the `EOC` tag missing. Took care of the typo.

### DIFF
--- a/src/ber/de.rs
+++ b/src/ber/de.rs
@@ -611,7 +611,11 @@ impl<'input> crate::Decoder for Decoder<'input> {
     }
 
     fn decode_optional<D: Decode>(&mut self) -> Result<Option<D>, Self::Error> {
-        self.decode_optional_with_tag(D::TAG)
+        if D::TAG == Tag::EOC {
+            Ok(D::decode(self).ok())
+        } else {
+            self.decode_optional_with_tag(D::TAG)
+        }
     }
 
     /// Decode an the optional value in a `SEQUENCE` or `SET` with `tag`.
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR!

PR #159, opened 2023-09-22
docs(README): add declaration reference

6d7a:
Added a set of data type examples to the README.


XAMPPRocky:
Thank you for your PR! I think if possible we should put this reference into the rust documentation directly. Another possibility would be making a mdbook for RASN, but for something like this I think having in the rustdoc directly would be the most beneficial.

6d7a:
@XAMPPRocky, I included the reference as a section in the README and corrected some errors.

README.md
```diff
@@ -212,6 +212,656 @@ struct Person2 {
...
+#[rasn(delegate)]
+struct TestTypeB(pub bool);
+
+/// either
+type TestTypeA = TestTypeB;
+
+/// or
+#[derive(AsnType, Decode, Encode)]
+#[rasn(delegate)]
+struct TestTypeA2(pub TestTypeB);
```
XAMPPRocky, line 247: One thing I'm not sure about here, is that you don't mention that just a type alias is sufficient. You don't need to have delegate types if there's no restrictions or you don't want/need to implement custom methods for it.

```rust
type TestTypeB = bool;
```

6d7a: True. I'll also add the possibility of type aliases in the other examples.

XAMPPRocky:
Thank you for your PR!

PR #162, opened 2023-09-27
Fix Result scoping

orthecreedence:
Fixes an instance where `Result` is used in the proc macros instead of `core::result::Result`. Without scoping, any crate-local types named `Result` confuse the compiler.


XAMPPRocky:
Thank you for your PR!

Issue #161, opened 2023-09-27
Newtypes with #[rasn(delegate)]


orthecreedence:
I'm seeing an issue with `#[rasn(delegate)]` which I'm not sure if it's expected or not. In short, delegate works on newtypes that wrap basic types, but not newtypes that wrap more complex types:

```rust
#[derive(Clone, Debug, rasn::AsnType, rasn::Encode, rasn::Decode)]
#[rasn(choice)]
enum Choose {
    #[rasn(tag(explicit(0)))]
    Single(String),
}

#[derive(Debug, rasn::AsnType, rasn::Encode, rasn::Decode)]
#[rasn(delegate)]
struct Nt1(Choose);

#[derive(Debug)]
struct Nt2(Choose);

#[derive(Debug, rasn::AsnType, rasn::Encode, rasn::Decode)]
#[rasn(delegate)]
struct Nt3(String);

impl rasn::AsnType for Nt2 {
    const TAG: rasn::Tag = rasn::Tag::EOC;
}

impl rasn::Encode for Nt2 {
    fn encode_with_tag<E: rasn::Encoder>(&self, encoder: &mut E, _tag: rasn::Tag) -> std::result::Result<(), E::Error> {
        self.0.encode(encoder)?;
        Ok(())
    }
}

impl rasn::Decode for Nt2 {
    fn decode_with_tag<D: rasn::Decoder>(decoder: &mut D, _tag: rasn::Tag) -> std::result::Result<Self, D::Error> {
        let inner = <Choose>::decode(decoder)?;
        Ok(Self(inner))
    }
}

let choose = Choose::Single("hi".into());
let str1 = "hi".to_string();
let nt1 = Nt1(choose.clone());
let nt2 = Nt2(choose.clone());
let nt3 = Nt3(str1.clone());

let ser_choose = rasn::der::encode(&choose).unwrap();
let ser_str1 = rasn::der::encode(&str1).unwrap();
let ser_nt1 = rasn::der::encode(&nt1).unwrap();
let ser_nt2 = rasn::der::encode(&nt2).unwrap();
let ser_nt3 = rasn::der::encode(&nt3).unwrap();

println!("Ch1 --- {:?}", ser_choose);
println!("Nt1 --- {:?}", ser_nt1);
println!("Nt2 --- {:?}", ser_nt2);
println!("St1 --- {:?}", ser_str1);
println!("Nt3 --- {:?}", ser_nt3);
```

The output of this is:

```
[...]

PR #163, opened 2023-09-28
Delegate newtype EOC

orthecreedence:
So, this one is probably controversial since it likely constitutes a breaking change. This is basically me scratching an itch with #161.

The scenario is explained in the linked issue, basically I want

```
Newtype3(Newtype2(Newtype1(ComplexType)))
```

where each of the `Newtype`s has `#[rasn(delegate)]` set in the derive macro to directly serialize as `ComplexType`. I added some tests with this PR as well.

If this doesn't get accepted I won't be hurt, but I would like to have some kind of way of doing the above because it's becoming very common for me to wrap some none-rasn-serializable value in another library with a `Newtype(SomeForeignType)` and I don't want to have to write a bunch of serialize/deserialize stuff by hand each time.

### DIFF
--- a/macros/src/decode.rs
+++ b/macros/src/decode.rs
@@ -27,11 +27,18 @@ pub fn derive_struct_impl(
             }
         } else {
             quote! {
-                <#ty as #crate_root::Decode>::decode_with_tag_and_constraints(
-                    decoder,
-                    tag,
-                    <#ty as #crate_root::AsnType>::CONSTRAINTS.override_constraints(constraints),
-                ).map(Self)
+                match tag {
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR! Looks good to me, you just need to resolve the merge conflict, and it should be good to merge.

> So, this one is probably controversial since it likely constitutes a breaking change.

A breaking change on its own is not controversial, namely we haven't released a major version yet, and secondly the only thing a major version has to do is show why it's worth making the change, and I think this PR shows that makes the behaviour more intuitive.

XAMPPRocky:
Thank you for your PR!

PR #166, opened 2023-10-03
Gensym field names

orthecreedence:
Use "gensym" idents when enumerating named fields. instead of raw field names (ie `name`, `tag`, etc), prefix the fields
with `__rasn_field_`: `__rasn_field_name`, `__rasn_field_tag`, etc.

This is so an enum like this can compile (it does not compile without this PR):

```rust
#[derive(AsnType, Decode, Encode)]
#[rasn(choice)]
enum MyEnum {
    #[rasn(tag(explicit(0)))]
    SomeVariant {
        #[rasn(tag(explicit(0)))]
	encoder: String,
        #[rasn(tag(explicit(1)))]
	tag: String,
    }
}
```

The idents `encoder` and `tag` conflict with the function args `encoder`/`tag` generated by the derive macro. In other words, it's not possible to have named enum variants of `encoder`, `tag`, or `constraints` because they conflict with the derived function definition.

This fixes things by prefixing the generated variable names.

### DIFF
--- a/macros/src/decode.rs
+++ b/macros/src/decode.rs
@@ -131,7 +131,7 @@ pub fn derive_struct_impl(
                 };
 
                 (
-                    quote!(const #const_name: Tag = #tag;),
+                    quote!(const #const_name: #crate_root::Tag = #tag;),
                     quote!((#context, #const_name) => { #choice_name::#field_name(#decode_impl) }),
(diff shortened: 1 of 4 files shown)


XAMPPRocky:
Thank you for your PR!

Issue #168, opened 2023-10-06
Nested choices not completly fixed on PER


Nicceboy:
Seems like the issue of infinite recursion was not completely fixed in https://github.com/XAMPPRocky/rasn/issues/148 and https://github.com/XAMPPRocky/rasn/pull/157 .

The fix only works if some parent type is struct or set.
In nested choices, the problem still occurs. 

I tried fixing it by modifying the function `encode_explicit_prefix` to following:
```rust
    fn encode_explicit_prefix<V: Encode>(
        &mut self,
        tag: Tag,
        value: &V,
    ) -> Result<Self::Ok, Self::Error> {
        if let Some((_, true)) = self.field_bitfield.get(&tag) {
            value.encode(self)
        } else if self.field_bitfield.get(&tag).is_none() {
            // There is no bitfield if none of the parent objects is struct/set
            // But we still need to handle nested choices explicitly
            value.encode(self)
        } else {
            self.set_bit(tag, true)?;
            value.encode_with_tag(self, tag)
        }
    }
``` 
In OER it works, but in PER it works only for two levels, and on third level the values are halved for some reason?

`asn1tools` example:

```python
import asn1tools


def main():
    # Compile the ASN.1 module
    spec = asn1tools.compile_string(
        """MyModule DEFINITIONS AUTOMATIC TAGS ::= BEGIN

       Choice ::= CHOICE {
           normal INTEGER,
           high   INTEGER,
           medium INTEGER
       }
       BoolChoice ::= CHOICE {
           a BOOLEAN,
           b BOOLEAN,
           c Choice
       }
       TripleChoice ::= CHOICE {
           a BOOLEAN,
           b BoolChoice
       }
       END""",
        codec="uper",
    )

    # Encode a value
    # value = {'c': {'normal': 333}}
    value = ("b", ("c", ("normal", 333)))
    encoded = spec.encode("TripleChoice", value)

    # Print the encoded value
    for e in encoded:
        print(f"{e}, ", end="")


if __name__ == "__main__":
    main()
```

Will produce `192, 16, 10, 10`.

On `rasn` encoding, the output is exactly half:
```console
thread 'uper::tests::only_nested_choice' panicked at 'assertion failed: `(left == right)`
Diff < left / right > :
 [
<    192,
<    16,
<    10,
<    104,
>    96,
>    8,
>    5,
>    52,
 ]
```
If I add one nesting level for `asn1tools`, then it produces the same output.
[...]

PR #169, opened 2023-10-06
Draft: Fix nested choice when no struct/set present

Nicceboy:
#168

### DIFF
--- a/src/per/enc.rs
+++ b/src/per/enc.rs
@@ -987,12 +987,16 @@ impl crate::Encoder for Encoder {
         tag: Tag,
         value: &V,
     ) -> Result<Self::Ok, Self::Error> {
-        if let Some((_,true)) = self.field_bitfield.get(&tag) {
+        if let Some((_, true)) = self.field_bitfield.get(&tag) {
+            value.encode(self)
+        } else if self.field_bitfield.get(&tag).is_none() {
+            // There is no bitfield if none of the parent objects is struct/set
+            // But we still need to handle nested choices explicitly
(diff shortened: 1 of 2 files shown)


Nicceboy:
Build seems to fail for third party issue:

```console
  = note: /usr/local/lib/gcc/x86_64-unknown-netbsd/9.4.0/../../../../x86_64-unknown-netbsd/bin/ld: cannot find -lexecinfo
          collect2: error: ld returned 1 exit status
          

error: could not compile `rasn-compiler` (bin "rasn-compiler") due to previous error
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

XAMPPRocky:
Thank you for your PR! I think the question I have is, is why is the variance wrong? If there's two variants, the variance should be two, not three. This PR solves this case, but doesn't really answer the root issue.

Nicceboy:
The variance is correct but the range for constraint is incorrectly set because index starts from zero (classic).

XAMPPRocky:
Ah gotcha, would you be able to make this a function? Having `- 1` peppered in the code is a bit of a smell as it's not obvious just by looking at it why, so it could be forgotten when the code is revisited.

Nicceboy:
I added separate variable declaration and some comments. Also debug asserts that variance length should be always greater than 0, which in theory should be always greater, but no other checks. (You cannot pass variant if enum does not have one...).
Would that be okay?

XAMPPRocky:
Thank you for your PR!

PR #172, opened 2023-10-10
Fix/issue 165

### LINKED ISSUE #165: Error when encoding/decoding non-explicitely tagged vecs
Using rasn v0.9.5. I recently decided to switch the serialization on the project I'm working on from `#[rasn(tag(explicit(N)))]` to `#[rasn(tag(N))` and I'm now having an issue with deserialization. I reduced my case down to two examples:

```rust
    #[test]
    fn ser_sequence_vec2() {
        #[derive(Debug, Clone, AsnType, Encode, Decode)]
        #[rasn(choice)]
        pub enum MultisigPolicySignature {
            #[rasn(tag(0))]
            Key {
                #[rasn(tag(0))]
                key: String,
            },
        }

        #[derive(AsnType, Clone, Debug, Decode, Encode)]
        struct Transaction {
            #[rasn(tag(0))]
            signatures: Vec<MultisigPolicySignature>,
        }

        #[derive(AsnType, Clone, Debug, Decode, Encode)]
        struct Transactions {
            #[rasn(tag(0))]
            transactions: Vec<Transaction>,
        }

        let transactions1 = Transactions {
            transactions: vec![
                Transaction {
                    signatures: vec![
                        MultisigPolicySignature::Key {
                            key: String::from("key"),
                        }
                    ],
                }
            ],
        };

        let ser1 = rasn::der::encode(&transactions1).unwrap();
        println!("ser1: {:?}", ser1);
        let transactions1_2: Transactions = rasn::der::decode(&ser1[..]).unwrap();
        println!("trans2: {:?}", transactions1_2);
        assert_eq!(transactions1_2.transactions.len(), 1);
    }
```

This test gives the output:

```
running 1 test
ser1: [48, 13, 160, 11, 48, 9, 160, 7, 48, 5, 128, 3, 107, 101, 121]
trans2: Transactions { transactions: [] }
thread 'util::ser::tests::ser_sequence_vec2' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', src/util/ser.rs:428:9
test util::ser::tests::ser_sequence_vec2 ... FAILED
```

Notice the `transactions []` ...it deserializes the outer vec as empty. This serializes to https://lapo.it/asn1js/#MA2gCzAJoAcwBYADa2V5 which I think is correct, so the issue might be in decoding.

This can be reduced even more:

```rust
    #[test]
    fn ser_sequence_vec2() {
        #[derive(Debug, Clone, AsnType, Encode, Decode)]
        #[rasn(choice)]
        pub enum MultisigPolicySignature {
            #[rasn(tag(0))]
            Key {
                #[rasn(tag(0))]
                key: String,
            },
        }

        #[derive(AsnType, Clone, Debug, Decode, Encode)]
        struct Transactions {
            #[rasn(tag(0))]
            transactions: Vec<MultisigPolicySignature>,
        }

        let transactions1 = Transactions {
            transactions: vec![
                MultisigPolicySignature::Key {
                    key: String::from("key"),
                },
            ],
        };

        let ser1 = rasn::der::encode(&transactions1).unwrap();
        println!("ser1: {:?}", ser1);
        let transactions1_2: Transactions = rasn::der::decode(&ser1[..]).unwrap();
        println!("trans2: {:?}", transactions1_2);
        assert_eq!(transactions1_2.transactions.len(), 1);
    }
```

Output:

```
running 1 test
ser1: [48, 9, 160, 7, 48, 5, 128, 3, 107, 101, 121]
thread 'util::ser::tests::ser_sequence_vec2' panicked at 'called `Result::unwrap()` on an `Err` value: FieldError { name: "Transactions.transactions", error: "Unexpected extra data found: length `7` bytes" }', src/util/ser.rs:416:74
test util::ser::tests::ser_sequence_vec2 ... FAILED
```

This serializes to https://lapo.it/asn1js/#MAmgBzAFgANrZXk which again looks correct to me.

Interestingly, this can be "fixed" in both examples by updating `MultisigPolicySignature` to either remove the tag specification or make it explicit:

```rust
        // works
        pub enum MultisigPolicySignature {
            Key {
                #[rasn(tag(0))]
                key: String,
            },
        }

        // works
        pub enum MultisigPolicySignature {
            #[rasn(tag(explicit(0)))]
            Key {
                #[rasn(tag(0))]
                key: String,
            },
        }
```

I'm guessing this has something to do with abiguity when dealing with SEQUENCE vs SEQUENCE OF, but not sure if there's an easy fix. I can try to pick at it and open a PR, but I am definitely moving out of my comfort space, so some guidance would be appreciated. Thank you!

6d7a:
Closes #165 

I cross-checked the encoding output in the issue using [asn1.io](https://asn1.io/asn1playground/), assuming the following ASN1 to represent the struct mentioned [here](https://github.com/XAMPPRocky/rasn/issues/165#issuecomment-1741602093):
```asn1
    Issue-165 DEFINITIONS IMPLICIT TAGS ::=
    BEGIN

    Signatures ::= SEQUENCE OF MultisigPolicySignature

    MultisigPolicySignature ::= CHOICE {
        key [0] SEQUENCE {
            key [0] UTF8String
        }
    }

    END
```
When encoding nested `SEQUENCE`s, the `rasn` encoder did not take the custom tag into account, which in case of the nestes `key`-`SEQUENCE` and was producing a wrong tag, i.e. `UNIVERSAL 0`. I changed the call in the macro for encoding inner items from a simple `encode` to `encode_with_tag`. The tests run fine, and I tried to trace the different cases we could encounter through the macro process and I couldn't find any side-effects on other cases. You'll likely have a much more complete view of how the different macro calls come together, @XAMPPRocky.

Let me know if I need to do more digging.

### DIFF
--- a/macros/src/encode.rs
+++ b/macros/src/encode.rs
@@ -153,11 +153,11 @@ pub fn map_to_inner_type(
         syn::Fields::Unit => (quote!(;), quote!()),
     };
 
+    let tag = tag.to_tokens(crate_root);
     let inner_impl = if is_explicit {
-        let tag = tag.to_tokens(crate_root);
         quote!(encoder.encode_explicit_prefix(#tag, &inner).map(drop))
     } else {
-        quote!(inner.encode(encoder))
+        quote!(inner.encode_with_tag(encoder, #tag))
     };
 
     quote! {
--- a/tests/issue165.rs
+++ b/tests/issue165.rs
@@ -1,0 +1,23 @@
+use rasn::{types::*, *};
+
+#[test]
+fn decode_implicitly_tagged_vec_choice() {
+    #[derive(Debug, Clone, AsnType, Encode, Decode, PartialEq)]
+        #[rasn(choice)]
+        pub enum MultisigPolicySignature {
+            #[rasn(tag(0))]
+            Key {
+                #[rasn(tag(0))]
+                key: String,
+            },
+        }
+
+    let transactions2 = vec![MultisigPolicySignature::Key {
+        key: String::from("key1"),
+    }];
+
+    let ser2 = rasn::der::encode(&transactions2).unwrap();
+    let transactions2_2: Vec<MultisigPolicySignature> = rasn::der::decode(&ser2[..]).unwrap();
+    assert_eq!(transactions2_2.len(), 1);
+    assert_eq!(transactions2_2.first().unwrap(), &MultisigPolicySignature::Key { key: String::from("key1") });
+}


XAMPPRocky:
Thank you for your PR!

orthecreedence:
Thanks @6d7a! I wasn't sure if it was an encode/decode issue...I was leaning decode, but looks like my intuition was wrong. Looks like I need to do a bit more reading of ASN1. I'll keep that asn1.io site in my back pocket from now on.

6d7a:
No worries, thanks for the extensive documentation of the issue, that helped a lot during debugging. asn1.io is a pretty solid resource, at least as far as 5 compilations per hour get you.

XAMPPRocky:
> at least as far as 5 compilations per hour get you.

Yeah let's just a say strong motivation for getting more of the codecs working and the compiler, is to hopefully one day create a version of asn1.io that can run on GitHub pages so you can compile and test ASN.1 without the ridiculous restrictions and licences OSS put on you.

PR #174, opened 2023-10-11
Fix/uper integer

6d7a:
If a value range can be determined for a given integer, `uper` does not handle values larger than 64K differently than those smaller than 64K.

#### ITU-T X.691 Reference
```
11.5.6	In the case of the UNALIGNED variant the value ("n" – "lb") shall be encoded 
as a non-negative binary integer in a bit field as specified in 11.3 with the minimum number 
of bits necessary to represent the range.
```


XAMPPRocky:
Thank you for your PR!

PR #176, opened 2023-10-11
fix(per): pad non-negative binary integer slice at start

6d7a:
The workaround `to_vec` function in `src/per` works fine in all cases but when decoding non-negative binary integers. To be read correctly into a `BigUint`, the bitslice needs to be padded at the start. I added an additional parameter to specify where padding should be inserted when converting the bitslice to `Vec<u8>`.

If you have a minute, @XAMPPRocky, could you let me know if there is a more elegant way of handling the two cases [in `to_vec`](https://github.com/6d7a/rasn/blob/8e275504e9c58d3398fc024f63e7c691ca24e6bc/src/per.rs#L78)?

Thanks

### DIFF
--- a/src/per.rs
+++ b/src/per.rs
@@ -68,13 +68,35 @@ pub(crate) fn range_from_bits(bits: u32) -> i128 {
 }
 
 // Workaround for https://github.com/ferrilab/bitvec/issues/228
-pub(crate) fn to_vec(slice: &bitvec::slice::BitSlice<u8, bitvec::order::Msb0>) -> Vec<u8> {
+pub(crate) fn to_vec(
+    slice: &bitvec::slice::BitSlice<u8, bitvec::order::Msb0>,
+    pad_start: bool,
+) -> Vec<u8> {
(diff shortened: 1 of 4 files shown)


XAMPPRocky:
> If you have a minute, @XAMPPRocky, could you let me know if there is a more elegant way of handling the two cases [in to_vec](https://github.com/6d7a/rasn/blob/8e275504e9c58d3398fc024f63e7c691ca24e6bc/src/per.rs#L78)?

If I'm reading the code right, there's only one spot where it's true and the rest are false, so I would just make a second function the pad_left case. There's no reason AFAICT that it has to be in to_vec.

6d7a:
Thanks, I was actually thinking more about the repeated for loop. I like the extra parameter in `to_vec`, because it makes it clear where padding will be added for all cases.

XAMPPRocky:
I'm still not sure I see the benefit to it being one function rather than two functions? I think two functions is clearer, looking at the callsite it's not clear what the boolean does.

PR #177, opened 2023-10-16
Fix: four uper issues

6d7a:
Closes #176 

This PR fixes four small (u)per issues:
* non-negative binary integer slices were padded right instead of left (I refactored #176 into two separate functions for the different padding use cases)
* extensible ENUMERATED without extended variants was treaded like non-extensible ENUMERATED
* contents of CHOICEs with only one option were not encoded
* extended INTEGERs were not encoded correctly

### DIFF
--- a/macros/src/enum.rs
+++ b/macros/src/enum.rs
@@ -86,6 +86,8 @@ impl Enum {
             }
         });
 
+        let extensible = self.config.constraints.extensible;
+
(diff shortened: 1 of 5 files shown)


XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR!

PR #179, opened 2023-10-17
fix(per): encoding extensible string types

6d7a:
This PR fixes a couple of issues with PER-encoding extensible string types, such as `BIT STRING`, `OCTET STRING` or `VisibleString`. Before, extensible string types were in some cases encoded with a duplicate extension marker, in some cases with a wrong length determinant.
I also refactored the string encoders so that extension bits are always written using the `encode_extensible_bit` function, and removed some code duplication by having `encode_length` call the `encode_string_length` function, which is largely identical.

### DIFF
--- a/src/per/enc.rs
+++ b/src/per/enc.rs
@@ -556,7 +556,7 @@ impl Encoder {
                     && size_constraint.constraint.contains(&octet_string_length)
             })
         });
-        let Some(constraints) = constraints.size() else {
+        let Some(size) = constraints.size() else {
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! Could you add some test cases for this, to prevent future regressions?

6d7a:
Sure, will do.

XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR!

Issue #109, opened 2022-11-01
Support for XER


ERnsTL:
* in addition to the binary formats, support for the text-based XML rules and JSON encoding rules would be great in order to use this library for web apps
* true power of ASN1 is the many rulesets -> add XER and JER?

Is this planned?

XAMPPRocky:
Thank you for your issue! Yes, I would definitely like to support this, first I need to complete my PER implementation, and then I intend to focus on adding text based encoding rules like JER and XER.

ERnsTL:
@XAMPPRocky Thanks for your positive response - you took the words out of my mouth, was to ask about JER as well.

That is a great outlook! JER and XER can be great bridges between more conservative ASN1 systems and hip JSON-based apps and the web.

Cool, and have a great day!

ERnsTL:
> JER is now implemented so only XER is left.

This is some of the best news in recent days! Thanks for your efforts

XAMPPRocky:
I didn't implement, it was a contribution from the community but I'll accept it regardless 😄

PR #187, opened 2023-10-27
Feature/jer

6d7a:
Adds support for [JSON encoding rules](https://obj-sys.com/docs/JSONEncodingRules.pdf)

JER-support is implemented as a non-default feature, both in `rasn` as the feature `jer`, as well as in `rasn-derive` as the feature `text-based-encodings` (changes in `rasn-derive` will likely be needed for XER handling, too). Encoder and decoder are built upon `serde_json`.

Closes #188 
#109


XAMPPRocky:
Thank you for your PR!

I have two initial bits of feedback, the first is that I don't think this should be a feature. If we want to go for that approach we should do that separately and apply it uniformly to the other codecs.

The second is the use of serde. Since we don't benefit from the generic codec framework of serde, since we are a codec framework ourselves, wouldn't we be better off using a purpose built json library, like jzon?

6d7a:
`jzon` looks indeed promising, I overlooked it  designing the JER implementation. The API is similar to `serde_json` so it won't take much effort ro refactor. Seeing that `jzon` doesn't have any further dependencies I don't see any harm in not making JER a feature. I'll follow up with an update.

macros/Cargo.toml
```diff
@@ -12,6 +12,8 @@ repository.workspace = true
 [lib]
 proc-macro = true
 
+[features]
```
XAMPPRocky, line 15: No longer needed

README.md
```diff
@@ -33,6 +33,7 @@ The encoder and decoder have been written in 100% safe Rust and fuzzed with [Ame
 - Distinguished Encoding Rules (DER)
 - Aligned Packed Encoding Rules (APER)
 - Unaligned Packed Encoding Rules (UPER)
+- JSON Encoding Rules (JSON) (cargo feature `"jer"`)
```
XAMPPRocky, line 36: No longer needs a feature

6d7a:
@XAMPPRocky I replaced `serde_json` with `jzon` and moved the `jer` feature into the main codebase. I also made a small change to `Codec` in order to handle text-based encoding rules. Unfortunately, some fuzz tests fail under specific builds and I'm having a hard time figuring out why. Do you have an idea? Thanks!

src/codec.rs
```diff
@@ -43,20 +49,78 @@ impl Codec {
             Self::Cer => crate::cer::encode(value),
             Self::Der => crate::der::encode(value),
             Self::Uper => crate::uper::encode(value),
+            Self::Jer => Err(crate::error::EncodeError::from_kind(
+                crate::error::EncodeErrorKind::Custom {
+                    msg: "JER is a text-based encoding. Call `Codec::encode_to_string` instead."
+                        .into(),
+                },
+                Codec::Jer,
+            )),
```
XAMPPRocky, line 58: Should this be the behaviour? Why not return the bytes of text based encoding. `encode_to_string`'s behaviour makes sense, but you can represent strings as bytes.

src/codec.rs
```diff
@@ -43,20 +49,78 @@ impl Codec {
...
             Self::Cer => crate::cer::decode(input),
             Self::Der => crate::der::decode(input),
             Self::Uper => crate::uper::decode(input),
+            Self::Jer => Err(crate::error::DecodeError::from_kind(
+                crate::error::DecodeErrorKind::Custom {
+                    msg: "JER is a text-based encoding. Call `Codec::decode_from_str` instead."
+                        .into(),
+                },
+                Codec::Jer,
+            )),
```
XAMPPRocky, line 84: Same as the above

src/jer/de.rs
```diff
@@ -0,0 +1,583 @@
...
+                )
+            })
+    }
+
+    fn decode_teletex_string(
+        &mut self,
+        _t: crate::Tag,
+        _c: Constraints,
+    ) -> Result<TeletexString, Self::Error> {
+        todo!()
```
XAMPPRocky, line 227: Why todo here?

src/jer/enc.rs
```diff
@@ -0,0 +1,414 @@
...
+        ))
+    }
+
+    fn encode_teletex_string(
+        &mut self,
+        _t: crate::Tag,
+        _c: crate::types::Constraints,
+        _value: &crate::types::TeletexString,
+    ) -> Result<Self::Ok, Self::Error> {
+        todo!()
```
XAMPPRocky, line 216: Same as above

XAMPPRocky (commented):
LGTM to me mostly, just a couple comments.

src/codec.rs
6d7a: You have a point. I made the appropriate changes.

src/jer/de.rs
6d7a: Currently we do not have any logic to convert between `String` and `TeletexString` or to validate a `TeletexString`'s bytes and it doesn't seem to be a trivial addition to make. PER also has a `todo!()` in its `Decoder` impl, so I'd propose to tackle this issue in a separate PR.

Nicceboy: Just a note when out of curiosity looking this PR, Teletex standard has been withdrawn. Should we just delete the type? https://www.itu.int/rec/T-REC-T.61 https://en.wikipedia.org/wiki/ITU_T.61 and focus on other more used ones. Get back to it if someone really needs it.

6d7a: Unfortunately [it keeps sticking around](https://www.itu.int/ITU-T/formal-language/itu-t/x/x509/2019/PkiPmiExternalDataTypes.html) 🥲  It is also used in `rasn-pkix`. However, I think we can afford to be lenient when it comes to the implementations in `jer`, `xer`, `oer`, or `per`, since AFAIK T.61 had already been revoked when they were conceived.

Nicceboy: Oh, I missed that one. It is mentioned in `oer` as (TeletexString, T61String) both, but I haven't faced any standard which actually uses the types yet.

Nicceboy: This might be good reference if we end up implementing it:

https://github.com/mouse07410/asn1c/blob/84d3a59c1bb89c59be6ca0625bb14ebea9084ba5/skeletons/TeletexString.c#L4

XAMPPRocky:
Thank you for your PR!

XAMPPRocky:
Thank you for your PR!

PR #188, opened 2023-10-27
fix(macros): handle negative discriminants

6d7a:
Implementing JER-support, I noticed that negative enum discriminators are not handled correctly in `macros`. This PR fixes that.


XAMPPRocky:
Looks good, but the CI isn't passing

6d7a:
@XAMPPRocky I see that the CI checks have passed. Could we merge this?

XAMPPRocky:
Thank you for your PR!

PR #190, opened 2023-10-30
Move some generally useful functions from PER to utility module

Nicceboy:
I think these functions (log2, to_vec etc.) might be better located on `utils.rs` or something similar, since they are not codec specific. For example, OER needs them as well.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,6 +41,7 @@ pub mod types;
 // Data Formats
 
 mod per;
+mod utils;
 
 pub mod aper;
 pub mod ber;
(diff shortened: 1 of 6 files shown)


XAMPPRocky:
Thank you for your PR! The only thing I would change is that I wouldn't have a single `utils` module. I don't like those because it's too generic, anything could be put there.

I would instead make two modules `num` and `bits`. I would move `log2` to `num`, the rest to `bits` and rename `range_from_bits` to `range_from_len`.

Nicceboy:
I made the change. Should these have own directory as well or current is good?

XAMPPRocky:
I don't think we need them to have a directory unless they have sub modules.

PR #195, opened 2023-11-06
Use `ok_or_else` instead of `ok_or` in Error handling for performance

Nicceboy:
Ref https://github.com/XAMPPRocky/rasn/discussions/194


XAMPPRocky:
Thank you for your PR!

PR #197, opened 2023-11-08
Field_error improved, Boxed error `kind`, explicit naming also for `DecodeErrorKind`

Nicceboy:
Boxed `Kind` type for error variants.
Also some other error improvements.

~~Parser errors from `Incomplete` and `MismatchedTag` does not include backtrace anymore.
Should restore the performance.~~

### DIFF
--- a/macros/src/config.rs
+++ b/macros/src/config.rs
@@ -1,4 +1,4 @@
-use std::ops::{Deref, Neg};
+use std::ops::Deref;
 
 use quote::ToTokens;
 use syn::{Lit, NestedMeta, Path, UnOp};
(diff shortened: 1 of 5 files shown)


XAMPPRocky:
Thank you for your PR! I'm unsure about making the backtrace optional. Since I'm not sure this is less complex than fixing the decode_sequence_of and decode_set_of

Nicceboy:
I was also thinking cases if someone builds logic for 'Incomplete' errors, not just the internal handling.

But I can try to find another solution.

Nicceboy:
I reverted the optional backtrace, maybe it is better to check other changes first

PR #206, opened 2023-11-22
Fix/issue 204

### LINKED ISSUE #204: Unable to round-trip NumericString with APER
So I originally was attempting to decode a PER `NumericString` and got weird results, so I wanted to compare with the `rasn` output for the same value, and encountered an encoding error. 

Here is a example which demonstrates the issue:

```
use rasn::AsnType;
use rasn::prelude::*;

#[derive(Debug, AsnType, rasn::Encode, rasn::Decode)]
#[rasn(delegate, from("0..=9"), size("1..=255"))]
struct SimpleNumericString(pub NumericString);

fn main() {
    let string = SimpleNumericString(NumericString::from_bytes(b"123451234512345").expect("Failed to construct NumericString"));
    println!("Original: {:?}", string);
    let encoded = rasn::aper::encode(&string).expect("Failed to encode");
    println!("Encoded: {:?}", encoded);
    let decoded: SimpleNumericString = rasn::aper::decode(&encoded).expect("Failed to decode");
    println!("Decoded: {:?}", decoded);
}
```

which crashes with

```
thread 'main' panicked at /home/fpr/projects/rasn/src/per/enc.rs:280:79:
range end index 15 out of range for slice of length 8
```

I did some initial investigation at the panic site [here](https://github.com/XAMPPRocky/rasn/blob/12e245e92d6b091035a175f59823477eca2f20a4/src/per/enc.rs#L278). At this point, `range` is `0..15`, which appears to be the related to the length of the original string, but `value` is `Some([18, 52, 81, 35, 69, 18, 52, 5])` which appears to be the packed string data, so the range is invalid to index the slice.

I was using `rasn` 0.12.1 and `rustc 1.76.0-nightly (2b603f95a 2023-11-12) ` from `nightly-x86_64-unknown-linux-gnu`.

6d7a:
Currently, `NumericString`s are not handled correctly when using `PER` codecs because the character indices in the numeric char set are not determined correctly. This PR fixes this behavior and adds corresponding test cases.

Closes #204


XAMPPRocky:
Thank you for your PR!

PR #208, opened 2023-11-25
Fix issue with decoding of BasicOcspResponse.

chrismoos:
Fixes an issue decoding a `BasicOcspResponse`, looks to be because the `version` field's `default` attribute was not being picked up.

Not sure if this was a typo or if there is a compiler that generates this and the issue is from there.


XAMPPRocky:
Thank you for your PR! The proc macro should have caught this.

PR #214, opened 2023-12-05
Make backtraces as feature, enabled by default

Nicceboy:
This PR makes backtraces optional, while enabled by default.

Sometimes it can be very desireable to disable them totally.

E.g. when fuzzing, with backtraces you get unstable 100-5000 execs per second, but without them it is 50k execes per second (from OER and with AFL).

There might be some other scenarios where performance gets priority over exact error sources.


XAMPPRocky:
Thank you for your PR!

PR #219, opened 2024-02-02
fix(macro): recognize option references

### LINKED ISSUE #217: Question about DER encoding discrepancy between versions (0.7 -> 0.8)
Hi! I had some tests for my project to make sure the serialization format is preserved across project updates. I was initially only testing *deserialization* but recently made a change in my project that made testing serialization deterministically possible as well. The interesting thing is that I noticed the serialization format changed from when I wrote the test (this was back on v0.7).

I wrote a quick test to check this out and came up with two results:

```rust
#[derive(Debug, Clone, AsnType, Encode, Decode)]
#[rasn(choice)]
enum Tx {
    #[rasn(tag(explicit(0)))]
    Test,
    #[rasn(tag(explicit(1)))]
    Reset {
        #[rasn(tag(explicit(0)))]
        keys: Option<Vec<u32>>,
    }
}
let tx1 = Tx::Reset { keys: None };
let tx1_ser = rasn::der::encode(&tx1).unwrap();
println!("--- {:?}", tx1_ser);
```

In v0.7 this produces (https://lapo.it/asn1js/#oQIwAA):
```
[161, 2, 48, 0]
```
And in v0.8 this produces (https://lapo.it/asn1js/#oQQwAqAA):
```
[161, 4, 48, 2, 160, 0]
```

I'm not entirely sure which one "correct" now. It seems v0.8+ encode `None` as their own sequence instead of omitting the value...is this the desired behavior? From my (admittedly limited) understanding, `OPTIONAL` fields are omitted when they are blank.

6d7a:
Closes #217 

`OptionalEnum::is_option_type` did not properly process references to `Option`s. References to `Option` types are present in enum values with named fields. This PR fixes that behavior and adds a test for it.

### DIFF
--- a/macros/src/config.rs
+++ b/macros/src/config.rs
@@ -421,6 +421,9 @@ impl OptionalEnum {
                 .segments
                 .last()
                 .map_or(false, |segment| segment.ident == self.path),
+            syn::Type::Reference(syn::TypeReference { elem, .. }) => {
+                self.is_option_type(elem)
+            }
             _ => false,
         }
     }
--- a/macros/src/config.rs
+++ b/macros/src/config.rs
@@ -421,9 +421,7 @@ impl OptionalEnum {
                 .segments
                 .last()
                 .map_or(false, |segment| segment.ident == self.path),
-            syn::Type::Reference(syn::TypeReference { elem, .. }) => {
-                self.is_option_type(elem)
-            }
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR!

orthecreedence:
@6d7a swoops in for another fix. Thanks for this!

PR #221, opened 2024-02-02
refactor: remove legacy compiler

6d7a:
Since we have the independent [compiler](https://github.com/librasn/compiler) now, can we remove the `compiler` member?


XAMPPRocky:
Thank you for your PR!

PR #223, opened 2024-02-12
fix(types): stack overflow in ObjectIdentifier PartialEq

### LINKED ISSUE #222: stack overflow when comparing &ObjectIdentifier with &[u32]
```
#[test]
fn test_stack_overflow() {
    let arr: &[u32] = &[1, 2, 3];
    let oid = ObjectIdentifier::new(arr).unwrap();
    if &oid != arr {
        assert!(false);
    }
    println!("done");
}
```
results in 
`fatal runtime error: stack overflow`
at least with rust-1.74.1

6d7a:
The implementation of PartialEq<[u32]> for ObjectIdentifier did not compare the inner Cow<[u32]> but 'self' with the other [u32]. This triggered an infinite ping pong between Deref::deref and PartialEq::eq.

Closes #222

### DIFF
--- a/src/types/oid.rs
+++ b/src/types/oid.rs
@@ -225,7 +225,7 @@ impl PartialEq<ObjectIdentifier> for &Oid {
 
 impl PartialEq<[u32]> for ObjectIdentifier {
     fn eq(&self, rhs: &[u32]) -> bool {
-        self == rhs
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR!

PR #227, opened 2024-03-06
refactor(macros): Treat Unit Structs as ASN.1 NULL

### LINKED ISSUE #225: UnbindRequest encodes/decodes incorrectly
Encoding `rasn-ldap::UnbindRequest` into the `ber` format results in `[0x62, 0]`, but the correct encoding should be `[0x42, 0]`. This also leads to decoding failures of correctly encoded messages.

The difference is in the 5th bit in tag (primitive/constructed). Since `UnbindRequest` request is a struct, it gets encoded as a sequence, with the bit set to 1 (constructed). Correctly it should be set to primitive, as it's just NULL with tag (https://datatracker.ietf.org/doc/html/rfc4511#section-4.3). 

I've checked the encoding against libldap based tools (e.g. ldapsearch).

6d7a:
Closes #225

Replaces the empty struct of `rasn_ldap::UnbindRequest` with a null delegate. @michaldrabina we use rust's unit value `()` as an equivalent for ASN.1's `NULL` value. That way, you can simply use the `delegate` annotation. I couldn't see that it breaks any existing code, but let me know in case. Thanks for the issue!

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,690 @@ struct Person2 {
 }
 ```
 
+## Reference
+The following table provides a range of examples showing how to declare data types with `rasn`.
+
+<table>
+<tr>
+<td></td> <td> ASN1 </td> <td> rasn </td>
+</tr>
+<tr>
+<td>Type alias</td>
+<td>
+
+```asn
+Test-type-b ::= BOOLEAN
+Test-type-a ::= Test-type-b
+```
+
+</td>
+<td>
+
+```rust
+// either
+use rasn::prelude::*;
+
+type TestTypeB = bool;
+
+type TestTypeA = TestTypeB;
+```
+```rust
+// or 
+use rasn::prelude::*;
+
+#[derive(AsnType, Decode, Encode)]
+#[rasn(delegate)]
+struct TestTypeB(pub bool);
+
+/// or
+#[derive(AsnType, Decode, Encode)]
+#[rasn(delegate)]
+struct TestTypeA(pub TestTypeB);
+```
+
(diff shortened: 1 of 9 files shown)


XAMPPRocky:
Thank you for your PR! Though I wonder if we should instead generate null from these kinds of structs instead of a delegate, I don't see where the empty sequences for unit structs are.

6d7a:
From the top of my head 😄 , [ISO8571-FTAM](https://www.itu.int/ITU-T/formal-language/iso/8571-4/1988/ISO8571-FTAM.html) contains some empty sequences. I guess it's better to be explicit with the null delegate.

XAMPPRocky:
Sure, but I guess I would consider the syntax for empty sequences to be the following, where it's defined to have fields, it's just the fields are empty.

```rust
struct FBeginGroupResponse {}
```

Where as in Rust the syntax we're using currently is called a "unit struct" meaning that semantically it is identical to `()`, but it is a better description of a single value.

```rust
struct UnbindRequest;
```

6d7a:
I see, makes sense. Should we move that refactor to a separate issue? By the way, do you think we can go ahead with PR #159?

XAMPPRocky:
> Should we move that refactor to a separate issue?

I'm not sure what you mean, fixing that should resolve the same issue right?

6d7a:
I adapted the macros so that unit structs are now interpreted as ASN.1 NULL.

XAMPPRocky:
Thank you for your PR!

PR #230, opened 2024-03-07
fix char and index maps, enable tests

### LINKED ISSUE #229: `index_map` and `character_map` are broken (one static map is initialized for all types, and thus doesn't work for different types)
As discussed in https://github.com/ferrilab/bitvec/issues/229#issuecomment-1984344301, the current `index_map` and `character_map` functions are broken:

https://github.com/librasn/rasn/blob/fb6e682d67caa8f3f751c66bf8067b8efc8c299f/src/types/strings/constrained.rs#L99-L112

The static `MAP: OnceBox<BTreeMap<u32, u32>>` exists once for all implementing types, not separately for each implementing type.

Trolldemorted:
- Fix #229 by having distinct OnceBoxes per implementing type
- Enable disabled tests


Trolldemorted:
I have fixed the formatting issues in the locations I touched. CI should still fail, because there are other locations where `cargo fmt` complains. Should I fix them too to make CI pass?

Trolldemorted:
rebased on current master, ran cargo fmt on everything

XAMPPRocky:

Trolldemorted:
I have noticed that the `long_sequence_of` test is disabled, but it passes on my machine™ (at least now). Was it disabled for this reason, and should be reenabled now as well?

XAMPPRocky:
If it passes it should be re-enabled, I forget why it was disabled.

PR #232, opened 2024-03-10
enable long_sequence_of

Trolldemorted:


XAMPPRocky:
I believe it's the amount of items exceeds the size of a pointer. The capacity is the size of a pointer on the platform.

Trolldemorted:
Do you have then any clue why that particular test would fail? Since we smell something fishy again, what could be our causes?

- A bug in that version of Rust beta on i686?
- A spurious bug that just happens once every few runs?
- A bug in rasn that occurs only if you execute it on a x86 machine and not on x64 with `cargo test --target i686-pc-windows-msvc`?

Could you restart the CI job and we observe whether this happens everytime?

XAMPPRocky:
I've restarted it now.

XAMPPRocky:
Seems like it was just spurious. Thank you for your PR!

Trolldemorted:
Could you fix the CI on the master branch? Then we'd have at least some history of when and how often things fail, right now almost every commit fails

PR #183, opened 2023-10-25
WIP: Add clippy and fix reported issues

wiktor-k:
Hi, I've been working on adding clippy to this project and most of the issues are minor but I'll need your help to resolve the remaining ones:

1. Using `to_*` name but not consuming `self`: This would require changing function signature:

```
warning: methods with the following characteristics: (`to_*` and `self` type is `Copy`) usually take `self` by value
  --> macros/src/tag.rs:37:22
   |
37 |     pub fn to_tokens(&self, crate_root: &syn::Path) -> proc_macro2::TokenStream {
   |                      ^^^^^
   = help: consider choosing a less ambiguous name
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
```

2. Interior mutability of OctetString:

```
warning: a `const` item should never be interior mutable
  --> standards/cms/src/authenticode.rs:31:1
   |
31 |   pub const SPC_CLASS_UUID: OctetString = OctetString::from_static(&[
   |   ^----
   |   |
   |  _make this a static item (maybe with lazy_static)
   | |
32 | |     0xa6, 0xb5, 0x86, 0xd5, 0xb4, 0xa1, 0x24, 0x66, 0xae, 0x05, 0xa2, 0x17, 0xda, 0x8e, 0x60, 0xd6,
33 | | ]);
   | |___^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
```

3. `Encode` adds mutable key type (which annoys Hash derive):

```
warning: mutable key type
  --> standards/cms/src/pkcs7_compat.rs:24:41
   |
24 | #[derive(AsnType, Clone, Debug, Decode, Encode, PartialEq, Eq, PartialOrd, Ord, Hash)]
   |                                         ^^^^^^ in this derive macro expansion
   |
  ::: /home/wiktor/src/open-source/rasn/macros/src/lib.rs:54:1
   |
54 | pub fn encode_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
   | ------------------------------------------------------------------------------- in this expansion of `#[derive(Encode)]`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#mutable_key_type
```

I'm not quite sure if some of them are false positives and I think you could comment on that.

(for the record I ran `cargo clippy --all` locally)

PR #234, opened 2024-03-13
Add clippy and fix reported issues

wiktor-k:
This PR replaces #183 since the code diverged and I actually mis-push'd :sweat_smile: 

It still leaves a couple of explicit `#[allow]`s for manual inspection :)


wiktor-k:
> (signal: 6, SIGABRT: process abort signal)

Hmm... The error looks weird. I can't reproduce it locally. Is this a spurious error?

XAMPPRocky:
Thank you for your PR!

PR #237, opened 2024-03-21
Manually fix the incompatible versions (a result of v0.13.0 release) and some clippy warnings

Nicceboy:
Updated the dependency versions.


Nicceboy:
Hmm, `clippy` warnings in CI seems to conflict with our best interests.

It wants to change `ok_or_else(` to `ok_or(`. This was performance issue with backtracing if I remember correctly.

PR #239, opened 2024-04-02
Feat/identifier annotation

6d7a:
This is related to [compiler issue #6](https://github.com/librasn/compiler/issues/6).

Text-based encoding rules use type names in their encoding of ASN.1 types. The type names in `rasn` bindings may not be the same as in the original spec. Certain type names - for example those that contain hyphens - are not acceptable as identifiers in rust, at all. This PR adds a new `identifier` annotation to retain the original names in `rasn` bindings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -724,6 +724,33 @@ struct TestTypeA {
 </td>
 </tr>
 <tr>
+<td>Renaming fields</td>
+<td>
+
+```asn
+Test-type-a ::= SEQUENCE { 
+    notQuiteRustCase INTEGER
+}
+```
+
+</td>
+<td>
+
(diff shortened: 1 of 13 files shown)


XAMPPRocky:
Thank you for your PR! This all looks good to me, but I don't see mention of why the JER codec is replacing hyphens with underscores, can you explain that part?

6d7a:
Sure. That conversion is a requirement of JER itself. From the JER spec:
_"An ASN.1 identifier is also a valid 
JavaScript identifier name if and only if it does not contain any hyphens. This makes the conversion rule simple: convert any hyphen into an underscore"_

PR #243, opened 2024-04-04
Fix/jer compliance

6d7a:
Turns out I was consulting the wrong spec for the JER implementation. This PR fixes encoding and decoding of `BIT STRING`s and removes erroneous identifier conversion. This way `jer` should be compliant with X.697 (with default encoding controls).

Thank you to @v0-e for pointing out issues with `jer` 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -724,6 +724,33 @@ struct TestTypeA {
 </td>
 </tr>
 <tr>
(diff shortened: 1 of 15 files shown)


XAMPPRocky:
Thank you for your PR!

PR #247, opened 2024-04-30
Remove `backtraces` from `rasn`'s `default` feature

### LINKED ISSUE #246: Standards crates (`rasn-pkix`, `rasn-ocsp`, etc.) should not enable `rasn`'s default feature
The standards crates enable `rasn`'s `default` feature, which can incur a significant performance penalty (due to the `backtraces` feature I assume).
These crates should depend on `rasn` with `default-features = false`, and turn on whatever features they need individually.

s-arash:
This feature can incur a significant performance penalty, and since it's more of a debugging tool, it should be disabled by default.

In my experiments, there was a ~15x perf improvement when decoding OCSP responses with this feature disabled.

Closes #246

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ chrono = { version = "0.4.27", default-features = false, features = ["alloc"] }
 bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 
 [features]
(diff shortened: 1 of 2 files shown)


XAMPPRocky:

PR #252, opened 2024-05-21
CI: Run Clippy for all targets

Nicceboy:
Maybe it is not bad thing if we run Clippy for all targets.
This also applies suggested fixes for tests.


XAMPPRocky:
Thank you for your PR!

Issue #244, opened 2024-04-10
Performance of PER codec


dudycz:
Hi. I have been playing with two most popular PER codecs: `rasn` and `asn1-codecs` and made some benchmark comparing their performance. One thing I have noticed is that rasn can be ~10x slower in encoding in some complex cases:

| Codec        | Encoding (µs)  | Decoding (µs)  |
|--------------|----------------|----------------|
| rasn         | 1916.7         | 826.8          |
| asn1-codecs  | 208            | 145.5          |

I looked into flamegraphs and callgrinds but I couldn't figure out what contributes to this big difference. If you're interested I could try to collect some and attach here.

Link to repo with benchmark: https://github.com/dudycz/asn1_codecs_bench

XAMPPRocky:
Thank you for your issue! Would you be able to attach the flamegraphs? That would be very helpful. In general the initial version I made prioritised correctness over performance so there's likely a lot of places where it could be more efficient. I already know that I haven't added the empty struct optimisation that was added to BER to PER.

dudycz:
I had to encode sample.asn 256 times to get something recorded in flamegraph. Here it is:
![rasn](https://github.com/librasn/rasn/assets/15647871/38bf1fc4-a6c1-4354-a1cb-3bc65adc94ca)

Nicceboy:
Seems like your test bench uses only integers. `asn1-codecs` crate seems to use `i128` internally for Integer type while `rasn` uses [BigInts](https://docs.rs/crate/num-bigint/latest). I would guess that this plays big factor at least. 

I think there was a plan to make inner implementation of integer types as feature, so that it would be possible to select different internal type for performance reasons if very big numbers are not required.

XAMPPRocky:
Yeah, that's still a todo, I should write up an issue giving details in case someone else who has more time is interested.

Nicceboy:
> Yeah, that's still a todo, I should write up an issue giving details in case someone else who has more time is interested.

I think this is rather important optimisation in general and should not take too much work to implement, if we have, for example, just the i128 and BigInts as initial options. If you have time to write what you had on your mind, maybe I can try to do it.

repnop:
@dudycz since #256 got merged a little while ago, are you able to rerun your benchmark, at least for the decoding side of things? it should be quite a bit better now.

dudycz:
Yes, I can confirm big improvement in decoding time (on my "bench" setup I observed decrease from 457us to 341us!). Good job!

[...]

Issue #76, opened 2022-03-04
Consider newtyping types instead of reexporting directly


repnop:
Instead of directly reexporting types such as `BigInt as Integer` and `Bytes as OctetString`, I think it makes a lot more sense to newtype these and provide a more stable interface so that the implementations can be changed without affecting the public API. For instance, being able to switch from using `num-bigint` to `ibig` for parsing variable-sized integers could provide a significant speed increase, avoiding unnecessary allocations for small integers. However, changing the `Integer` alias to `IBig` would be a breaking change. Reexporting types also makes the docs quite a bit more confusing in my opinion, as the documentation for the types ends up using a completely, totally different type name, for example when I go to view the docs for `OctetString`, I'm met with the following docs:
> A cheaply cloneable and sliceable chunk of contiguous memory.
> 
> `Bytes` is an efficient container for storing and operating on contiguous slices of memory. It is intended for use primarily in networking code, but could have applications elsewhere as well.
> 
> `Bytes` values facilitate zero-copy network programming by allowing multiple `Bytes` objects to point to the same underlying memory.

Which is a bit jarring when the type signature is named something completely different IMO

XAMPPRocky:
Thank you for your issue! This is definitelt planned, as I'm currently working on implementing PER and constraints, and constraints I want to eventually move the constraints impl to being completely `const` based, which requires having our own types to be able to add const parameters.

>  However, changing the `Integer` alias to `IBig` would be a breaking change.

I do want to be clear however that this is acceptable. This crate is still very firmly in `0.x` phase of development. So preventing breakage is not priority compared to iterating and improving rasn's features and performance.

> Which is a bit jarring when the type signature is named something completely different IMO

Yes, seperately I'd like to be able to massively revamp documentation, to provide more both informative and normative information on ASN.1, such that rasn's documentation is the primaary if not only source you need to use ASN.1 successfully.

Nicceboy:
I think I could start working with this [newtype support] after I finish OER codec, if that is okay.

I am still quite new for Rust, so I might need some design help.

But the basic idea would be to use Crate level features and with that to change types?

For the higher type (Integer) ans user facing API, we write our own functions, and then with crate features possibly map to different inner types.

XAMPPRocky:
> I am still quite new for Rust, so I might need some design help.

Right now we're mostly limited by Rust's const capabilities, once we have the ability to use enums it will make the design much more intuitive and straightforward.

[...]

PR #254, opened 2024-06-03
Rework Integer type for speed

Nicceboy:
Hi,

I have been working on with improving Integer type to get better performance.
Partial solution for https://github.com/librasn/rasn/issues/76

The current idea is make `Integer` type as enum as follows, after trying out many different approaches:

```rust
#[cfg(not(feature = "i128"))]
pub type PrimitiveInteger = i64;

#[cfg(all(target_has_atomic = "128", feature = "i128"))]
pub type PrimitiveInteger = i128;

pub const PRIMITIVE_BYTE_WIDTH: usize = core::mem::size_of::<PrimitiveInteger>()

#[derive(Debug, Clone, Ord, Hash, Eq, PartialEq, PartialOrd)]
#[non_exhaustive]
pub enum Integer {
    Primitive(PrimitiveInteger),
    Big(BigInt),
}
```

It uses the primitive type by default internally whenever possible to get the best performance, excluding cases where Big integer is used manually. Also, underflows or overflows are automatically converted to Big integer. `BigInt` could be easily changed to another type as well, if we would like to.

I am not sure if I over-engineered this one, but I wanted to make it easier to drop the mandated `i128` type requirement in the future if decided, so that `rasn` could be used on targets which do not support that type, as it is `#![no_std]` crate. 

Now, all the internal types can be changed just by changing `PrimitiveInteger` constant.

 * Removed most hard-coded casts from codecs to use `PrimitiveInteger` instead of `i128` when working with constraints. Needs more work to remove the type completely. 
 * Constraint types are tied to `PrimitiveInteger`, so `i128` is dropped from there.
   * Note for someone debugging constraints: constraint range is calculated silently and might wrap or saturated (e.g. range of `i64:MIN` and `i64:MAX` is 2x `i64:MAX` for UPER, and with `i64` type this does not fit and will show zero range. Not sure how unexpected behavior this is, but it is more likely than with `i128`, if someone does not use `i128`.

Overall, BER and OER will get around 2x performance boost for Integers. UPER will get less, as the are many other allocations in the code, which contribute more for the current performance.
OER can also get some significant boosts when allocations are optimized further.  

On M2 Pro, overall improvement based on benchmark mentioned in issue #244, when `i128` default used: 

image

With `i64` encoding is slightly faster, but decoding slower. 

I tried different styles, but enum one seem to get the best from both and generics are not exposed to any API. 

**The downside for this approach is, that when decoding data, it is not always certain what is the inner `Integer` type (Primitive or Big), if type is not constrained.** 

I implemented `Add` `Sub` and `Mul` operations for Integer, because it was easier for some codecs. Maybe it is better to not increase the amount of them, and instead encourage to use internal type ops instead. That would also add maintenance burden.

Current code will need next release of Rust (1.79, will be released on 13. June) to work. Otherwise, either one line of unsafe code or more complex code is required to bypass borrow checker.  

Also, it seems that it is not easy to automatically test the `i64` type by removing `i128` feature, because `rasn-pkix` is dev-dependency on the workspace, and it makes `rasn` tests compile with default features, no matter what (unless removed, or features changed for that crate). Maybe we don't need that bench with pkix? 

Maybe @6d7a also wants to take look since it is kinda big change.


XAMPPRocky:
Thank you for your PR! I don't know if you saw this discussion the other week, but I've an alternative proposal for improving integer decode speeds.

https://github.com/librasn/rasn/discussions/107#discussioncomment-9522935

I think this will provide better performance for most cases, as it will allow for the use of smaller container types.

I think we can use something like this though for codec internal numerics like length and similar.

Nicceboy:
> Thank you for your PR! I don't know if you saw this discussion the other week, but I've an alternative proposal for improving integer decode speeds.
> 
> [#107 (reply in thread)](https://github.com/orgs/librasn/discussions/107#discussioncomment-9522935)
> 
> I think this will provide better performance for most cases, as it will allow for the use of smaller container types without the need for branching since we know what integer type the user wants and can error out if bytes exceed the integer size.
> 
> I think we can use something like this though for codec internal numerics like length and similar.

Yeah, that will reduce branching. If I understand it correctly, we would need to rework the integer decode logic for every binary codec to get the best out of it? It is a bit more complex to add.

We have comparison point now to see the performance impact. However, it would be easier to see if we also optimize other allocations.

How about the encoding side? I assume it should be based on the trait as well. I tried one trait based approach earlier, but there the main challenge is to provide same output type for encoding. Single `to_vec` call for primitives will remove all the performance benefits. Big numbers are vectors and primitives are fixed-size slices (or pointers with unsafe). 

Since there is need to branch the logic based on fixed-size slice vs dynamic vector, I ended up using enum where it was easier. 

I can try to continue towards that suggestion, unless @repnop has already got started.

Nicceboy:
> Yes but it should only be changing a couple of lines in each, as AFAIK all of the codecs eventually get to the point where an integer is a bag of big endian bytes and converts it, so it would be just changing that part to use the trait.

I was thinking that maybe we could avoid some of that code before that point, since the constraints are passed, and if for every codec, we define own implementation for different types, we could hardcode the ranges and format for encodings in that specific constraint. At least in OER (probably it was bad design), it always calculates everything based on the constraint ranges. So some calculations could be potentially skipped (if that is even worth it). 

> Since we're using static dispatch, we can use impl Trait here to say it returns AsRef<[u8]> and we don't have care about what type we get back, because we just care about getting a &[u8].

Wow, that was too easy and seems to work well. I need to read more books...

Nicceboy:
Closing in favor of #256

PR #255, opened 2024-06-06
Add OID for MGF1

treebly:
Hi there! 

This is just a simple PR for adding the OID for the MGF1 hash function. AWS uses this in some of their KMS calls so it would be useful for it to be included here!

Relevant RFC: https://www.rfc-editor.org/rfc/rfc8017.html#appendix-B.2.1


XAMPPRocky:

PR #257, opened 2024-06-10
fix: decoding integers panic

repnop:
decoding integers encoded with byte lengths greater than `{integer}::BITS / 8` would cause a subtraction underflow and subsequent index panic


XAMPPRocky:
Thank you for your PR!

PR #263, opened 2024-06-14
perf(ber): don't allocate `2 + N` `BigInt`s when parsing OIDs

repnop:
the `BigInt`s that are used to parse OIDs are downcasted to `u32`s later on in the code anyway, so now that we have fixed-sized integer parsing that doesn't roundtrip through `BigInt`s, we can just attempt to deserialize them as `u32`s instead of allocating a bunch. this bought me a ~(-72%) benchmark win on parsing an SNMP packet since every varbind value has an associated OID, and got the parse time below `netsnmp`'s PDU decode for my microbenchmark :)

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -116,15 +116,15 @@ pub fn encode_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream 
 /// ##### Shared Attributes
 /// These attributes are available on containers, variants, and fields.
 /// - *`tag([class], number)`* — override the default tag with the one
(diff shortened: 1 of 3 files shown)


repnop:
looks like clippy got some new lints for doc comment formatting, so I added that in as well just so it won't be a problem for future PRs, if that's okay. I can split it off in its ownPR but figured since it was like ~3-4 lines of changes it'd be fine to just sneak in with this

XAMPPRocky:
Thank you for your PR!

Issue #262, opened 2024-06-14
Does not work in no-std environment


koxu1996:
I am unable to compile my project due to `found duplicate lang item panic_impl`, which appears to be caused by `std` library being indirectly included by `rasn` :fearful:. Even though this library is claimed as `no_std` its dependencies are not!

## Where `std` gets included?

Latest rasn v0.8.2 [depends](https://github.com/librasn/rasn/blob/1809d7c298ae3ac62da42749692e418f709d7147/Cargo.toml#L55) on **custom** nom parser with bitvec:

```
nom-bitvec = { package = "bitvec-nom2", version = "0.2.0" }
```

This need to be discussed. Why is it custom? Probably because upstream is not supporting `bitvec v1` - see https://github.com/rust-bakery/nom-bitvec/issues/1, and [relevant PR](https://github.com/rust-bakery/nom-bitvec/pull/2) got stuck. It seems that @XAMPPRocky released fork as renamed package with tag 0.2.0. The different between fork source code and files published in crates.io made my investigation more difficult.

Okay, so let's use crates.io as the source of truth. You can see that `bitvec-nom2` depends on `bitvec v1`:

![image](https://github.com/librasn/rasn/assets/6506780/19b5b473-cab6-49fc-9bd7-ed2c61b6eadf)

Default features are not disabled, so `std` gets included:

![image](https://github.com/librasn/rasn/assets/6506780/36ee0be8-c9d9-43bb-b05b-b7dcb0686c46)

**NOTE:** I was able to track it down with `cargo tree -e features`.

koxu1996:
The second place where `std` is indirectly included in `rasn`:

https://github.com/librasn/rasn/blob/1809d7c298ae3ac62da42749692e418f709d7147/Cargo.toml#L49

You can see that snafu's `backtraces` feature is enabling it:

![image](https://github.com/librasn/rasn/assets/6506780/d647b8ea-c818-41c3-8338-2e9a656e3ed5)

**NOTE:** `snafu 0.8.0` removed this feature

XAMPPRocky:
Thank you for your issue! This is definitely not intentional, we should disable std in my fork and put backtraces behind a std feature.

koxu1996:
Also JER should be put behind a feature, as `jzon` is not no-std compatible as well.

XAMPPRocky:
@koxu1996 Can you check now, as we now should compile with the latest nom-bitvec

Nicceboy:
[...]

PR #265, opened 2024-06-14
Hide `snafu` backtraces behind `std`

koxu1996:
One step toward `no_std` compatibility - see https://github.com/librasn/rasn/issues/262#issuecomment-2167887631.


XAMPPRocky:
Thank you for your PR!

PR #266, opened 2024-06-15
Hide `jer` behind feature (implying `std`)

koxu1996:
Quick attempt to hide JER implementation that is based on `jzon`, which requires `std` - see https://github.com/librasn/rasn/issues/262.


XAMPPRocky:

PR #268, opened 2024-06-18
COER: Fix invalid handling of zero padding on length of length

Nicceboy:
Validation was previously done based on the wrong variable.


XAMPPRocky:
Thank you for your PR!

PR #273, opened 2024-06-21
OER: decode error when unused preamble bits are not zero

Nicceboy:
Adds an error if unused bits on preamble are not zero.


XAMPPRocky:
Thank you for your PR!

XAMPPRocky:
This now needs to be rebased to resolve conflicts.

Issue #272, opened 2024-06-20
Combination of optional fields and explicit tag ignores the option type


Nicceboy:
When explicit tags are used for optional fields, wrong decoding (maybe encoding too) functions are derived. Or maybe option type disappears somehow, as the code does not hit the `decode_optional_*` methods. 

This seemed to be rather hard to debug, if there are any hints where this check happens.

```rust
use rasn::{oer, prelude::*};

#[derive(AsnType, Decode, Encode, Clone, Debug, PartialEq, Eq)]
pub struct SequenceOptionals {
    #[rasn(tag(explicit(0)))]
    pub it: Integer,
    #[rasn(tag(explicit(1)))]
    pub is: Option<OctetString>,
    #[rasn(tag(explicit(2)))]
    pub late: Option<Integer>,
}
fn main() {
    let test_seq = SequenceOptionals {
        it: 42.into(),
        is: None,
        late: None,
    };
    let encoded = oer::encode(&test_seq).unwrap();
    dbg!(&encoded);
    let decoded: SequenceOptionals = oer::decode(&encoded).unwrap();
    dbg!(&decoded);
    assert_eq!(test_seq, decoded);
}
```
Output (decoder does not hit the `decode_option_*`

```cmd
[testing/src/main.rs:19:5] &encoded = [
    0,
    1,
    42,
]
[testing/src/main.rs:21:5] &decoded = SequenceOptionals {
    it: 42,
    is: Some(
        b"",
    ),
    late: None,
}
thread 'main' panicked at testing/src/main.rs:22:5:
assertion `left == right` failed
  left: SequenceOptionals { it: 42, is: None, late: None }
 right: SequenceOptionals { it: 42, is: Some(b""), late: None }
stack backtrace:
   0: rust_begin_unwind
             at /rustc/ada5e2c7b5427a591e30baeeee2698a5eb6db0bd/library/std/src/panicking.rs:652:5
   1: core::panicking::panic_fmt
             at /rustc/ada5e2c7b5427a591e30baeeee2698a5eb6db0bd/library/core/src/panicking.rs:72:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/ada5e2c7b5427a591e30baeeee2698a5eb6db0bd/library/core/src/panicking.rs:363:5
   4: testing::main
             at ./src/main.rs:22:5
   5: core::ops::function::FnOnce::call_once
[...]

PR #274, opened 2024-06-21
OER: fix empty length decoding, no zero padding allowed

Nicceboy:
There was a logic bug when parsing zero padding and there was no length.

That was a cause for null string in #272 instead of None. 

But after consulting the standard again, I think zero padding is not allowed even in OER in this case and whole function can be removed. 
Zeros are allowed in long format.


XAMPPRocky:
This now needs to be rebased to resolve conflicts.

XAMPPRocky:
Thank you for your PR!

PR #276, opened 2024-06-22
Pass constraints correctly when encoding optional type

Nicceboy:
Constraints where not correctly passed in general when encoding optional type.


XAMPPRocky:
Thank you for your PR!

PR #277, opened 2024-06-23
OER: correctly encode null in option in extensions

Nicceboy:
A bit corner case, but fixes encoding of null value in optional extended type. 

I hope that I understood the use of `()` correctly.

The test should represent

```asn
Omitted ::= SEQUENCE { 
  a OctetString OPTIONAL,
  ...,
  omitted     NULL OPTIONAL
}
```

### DIFF
--- a/src/coer.rs
+++ b/src/coer.rs
@@ -1222,4 +1222,24 @@ mod tests {
             &[0x80, 0x01, 0x02, 0x03]
         );
     }
+    #[test]
+    fn test_null_in_option() {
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR!

Nicceboy:
I think this bug is also present in PER. The logic was copied from there initially.

XAMPPRocky:
@Nicceboy do you want to make a separate issue or PR for that? Just so it's tracked, as I don't know when I'll be able to get to it.

PR #278, opened 2024-06-24
PER: fix encoding of empty extension values

Nicceboy:
Related to the previous PR.

Actually this fixes all cases where encoding values would be empty in extensions, but the extension presence must be still part of the encoding.


XAMPPRocky:
Thank you for your PR!

PR #285, opened 2024-07-16
Fix error name regression introduced in a75b26b

kurkpitaine:
This is fix for a regression introduced in a75b26b, where `extension required but not present` was changed to `extension_present_but_not_required`, which is not the same meaning.

This fix proposes `required_extension_not_present` name to switch back and enforce the error meaning.


XAMPPRocky:

Issue #286, opened 2024-08-02
uper constraints not validated


shahn:
Hi, and thanks for rasn! I am wondering if my expectations are wrong or if there is a missing feature/bug. I am using rasn-compiler to generate a struct. I would have expected that the call to `encode()` would return an error, but too large values are accepted. In addition, depending on how much too large the value is, a roundtrip might still succeed with `decode()` also not checking the constraint.

```
#[derive(AsnType, Debug, Clone, Decode, Encode, PartialEq, PartialOrd, Eq, Ord, Hash)]
#[rasn(delegate, value("-500..=504"))]
pub struct Test(pub i16);

#[test]
fn it_works() {
    let x = Test(505);
    let output = uper::encode(&x).unwrap(); // I would have expected this to fail
    let y: Test = uper::decode(&output).unwrap();
    assert_eq!(x, y); // This assert does not trigger

    let x = Test(815);
    let output = uper::encode(&x).unwrap();
    let y: Test = uper::decode(&output).unwrap();
    assert_eq!(x, y); // This assert triggers with 815 != -209
}
```

In addition, it is possible to trigger a debug_assert inside rasn with this code:
```
    let x = Test(-501);
    let output = uper::encode(&x).unwrap();
```

It fails here:
```
panicked at src/types/constraints.rs:432:13:
assertion failed: value >= start
```

Am I somehow fundamentally misunderstanding the intended API, or can I help in adding the functionality if it is missing currently?

shahn:
I tried to add the validation for Integers, maybe that's a start? I think it is sufficient for my usecase for now

Nicceboy:
Hey, I checked most versions back to the version when UPER was added, and there seems to be no check. Encoding has never raised an error if it is out of range; I guess constraint is just used to encode it "correctly". However, it is not always correct if the value is outside of the range too much. Decoder tries to then decode based on correct range, and there is a mismatch. OER currently raises error when value is out of range, so I think it is okay add for PER too.

shahn:
Thanks for the feedback, I came to a similar conclusion which is why I wrote the patch :) I hope it is acceptable.

PR #287, opened 2024-08-02
Add integer value constraint checking for PER encoding

shahn:
This fixes issue #286.


shahn:
The failures look like clippy warnings that I did not introduce, but I have added a commit to fix them

shahn:
I am honestly a bit confused about these CI failures, now I see a failure in rasn-pkix for example. I will work on fixing them as well, but I did not touch any code over there.

XAMPPRocky:
I think the CI should probably be adjusted to have clippy only run on stable.

shahn:
I hope I did it right above :)

XAMPPRocky:

PR #290, opened 2024-08-05
Encoder: call correct method for Default and Optional types when value is None

kurkpitaine:
The algorithm introduced in 413a39a relies on the value's tag and an index incremented each time a value is processed. 
In sequences with mixed required and optional values, optional values were not set present in the preamble, causing decoding issues, because of calling `encode_none()` instead of `encode_none_with_tag()` in relevant cases.
This pull request fix this, by calling the correct encoding method for Default/Optional types when the value is None.


XAMPPRocky:
Thank you for your PR!

PR #299, opened 2024-08-22
Move BER integer leading byte handling to codec level

Nicceboy:
Integer decoding optimization PR added check for leading bytes when decoding Integer type, however this should happen on codec level rather than on type level.


XAMPPRocky:
Thank you for your PR!

PR #300, opened 2024-08-23
Remove `to_vec` and `to_left_padded_vec` as unnecessary and reduce allocations

Nicceboy:
I wanted to understand what is wrong with `bitslice.to_bitvec().into_vec()` call combination, because our custom functions add some overhead.

It might actually work as intended, see my comment [here.](https://github.com/ferrilab/bitvec/issues/228#issuecomment-2305929792)

For example, the original problem
```rust
assert_eq!(bitvec::bits![u8, Msb0; 1, 1, 0, 0, 0, 1, 1, 1, 0, 1][2..].to_bitvec().into_vec(), vec![29]);
```

This results to correct `BitVec` but the internal memory layout is not unified as result of slicing. Slicing here combines a set of references internally to construct the intended BitSlice (e.g. two first bits are omitted from the first byte). When we use `to_bitvec()`, this  internal layout is carried on, and internal byte alignment is the same, and `into_vec` result might be unexpected.  

To get the expected, unified memory layout, we need to call `.force_align()` to unify it after `to_bitvec()` call.

Then `.into_vec` will result into "expected" byte array, as `.into_vec` is based on memory presentation.

This seems to be noted on documentation, so maybe it is expected behavior. 

If we use the above correctly, we can drop all custom functions and extra allocations.

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -35,46 +35,6 @@ pub(crate) fn octet_string_ascending(a: &Vec<u8>, b: &Vec<u8>) -> Ordering {
     a.len().cmp(&b.len())
 }
 
-// Workaround for https://github.com/ferrilab/bitvec/issues/228
-pub(crate) fn to_vec(slice: &bitvec::slice::BitSlice<u8, bitvec::order::Msb0>) -> Vec<u8> {
-    use bitvec::prelude::*;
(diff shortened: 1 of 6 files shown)


XAMPPRocky:
Thank you for your PR!

PR #311, opened 2024-09-07
Reduce cloning in constraints

Nicceboy:
Small change, but not insignificant!

Function `override_constraints` could be optimized further later on, because it is one of the most used ones.  

image


XAMPPRocky:
Thank you for your PR!

PR #313, opened 2024-09-07
Add two custom profiles for benching lto and debugging

Nicceboy:
With some basic release build optimizations, we can get around 10-20% performance increase.
Maybe it is okay by default?

E.g. that integer bench:

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,16 @@ std = []
 backtraces = ["std", "snafu/backtraces"]
 jer = ["std", "jzon"]
 
+[profile.bench-lto]
+inherits = "bench"
+opt-level = 3
+codegen-units = 1
+lto = "fat"
+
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! But since this is library this wouldn't be passed on to downstream users, because our settings would be ignored and the application settings would be what's happening.

So I'm not sure it makes sense to benchmark with these enabled because someone has to enable these settings themselves to see the same benefit.

I think it might make sense for this to be a separate release profile so that we can benchmark it and compare it to default release.

Nicceboy:
Oh yes, that is true. I can change it to some benchmarks since that would be useful.

Nicceboy:
I changed it to be two profiles, one for lto and one for debugging. Other thing that I thought, was not possible, but it is not actually needed.

PR #314, opened 2024-09-08
Replace BTreeMap with HashMap in field_bitfield

Nicceboy:
`BTreeMap` is not the best type for its current use case in `field_bitfied`. 

We could change it to standard `HashMap` as easy improvement, which is very noticeable after the other allocations are optimized. In the future, this could be improved further as it should be possible to know the field count in compile-time and use stack instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "ahash"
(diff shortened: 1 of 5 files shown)


XAMPPRocky:
Thank you for your PR! Benchmark results please :)

Nicceboy:
It is mainly visible for OER codec currently, but it shares the same logic with PER:

image

It becomes easier to spot, after we make some easy allocation optimizations (e.g. reusing the same buffer)

For example, here is OER flamegraph for integer after reducing other allocations:

image

HashMap uses around half the time by using `fold`:

image

Using `heapless::LinearMap` would be even faster if we could provide the constant (tested by setting manual value)

image

PR #319, opened 2024-09-12
Clippy fixes and CI artifact upload/download update

Nicceboy:
Includes also fixes from the beta channel. There were also some unused types in tests.

### DIFF
--- a/src/error/decode.rs
+++ b/src/error/decode.rs
@@ -341,9 +341,12 @@ impl DecodeError {
     fn from_codec_kind(inner: CodecDecodeError) -> Self {
         let codec = match inner {
             CodecDecodeError::Ber(_) => crate::Codec::Ber,
+            #[allow(unreachable_patterns)]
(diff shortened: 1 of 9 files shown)


src/error/decode.rs
```diff
@@ -341,9 +341,12 @@ impl DecodeError {
     fn from_codec_kind(inner: CodecDecodeError) -> Self {
         let codec = match inner {
             CodecDecodeError::Ber(_) => crate::Codec::Ber,
+            #[allow(unreachable_patterns)]
```
XAMPPRocky, line 344: How are these unreachable?

XAMPPRocky (approved):

src/error/decode.rs
Nicceboy: These types are currently empty so they never have value, for example

```rust
#[derive(Snafu, Debug)]
#[snafu(visibility(pub))]
#[non_exhaustive]
pub enum CerDecodeErrorKind {}
```

PR #321, opened 2024-09-13
Constraint utilities, constraint const default, more const functions

Nicceboy:
Smaller update based on #318 which might be useful anyway already.

* Add ZERO associated constant for `IntegerType`
* Change Constraints::default() to be const
* Change some other constraint-related functions to be constant
* Add `VARIANCE_CONSTRAINT` for choice type - variance can be calculated in compile time
* Add macros to generate constant constraints, like

```rust
const CONSTRAINTS: Constraints = constraints!(value_constraint!(0, 100), size_constraint!(0, 100));
```

### DIFF
--- a/src/types/constraints.rs
+++ b/src/types/constraints.rs
@@ -2,7 +2,9 @@ use super::IntegerType;
 use alloc::borrow::Cow;
 use num_bigint::BigInt;
 
-#[derive(Debug, Default, Clone)]
+pub const DEFAULT_CONSTRAINTS: Constraints = Constraints::NONE;
+
+#[derive(Debug, Clone)]
 pub struct Constraints<'constraint>(pub Cow<'constraint, [Constraint]>);
 
 impl<'r> Constraints<'r> {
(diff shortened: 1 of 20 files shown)


Nicceboy:
It will also improve performance slightly, based on the stable compiler and default release build:

image

src/types/constraints.rs
```diff
@@ -384,8 +389,11 @@ impl<T: Default + Copy> Bounded<T> {
         }
     }
 
-    pub fn minimum(&self) -> T {
-        self.as_minimum().copied().unwrap_or_default()
+    pub const fn minimum(&self) -> T {
+        match self.as_minimum() {
+            Some(value) => *value,
+            None => T::ZERO,
```
XAMPPRocky, line 395: Isn't the minimum for signed integers below zero?

Nicceboy: It follows the existing default behavior, e.g. `isize` default is 0. I am not however sure should it be actually different.

PR #323, opened 2024-09-14
Add basic optimizations for allocations in OER

Nicceboy:
Lack of pre-allocations and not using existing buffers created many unnecessary allocations.

This will reduces them a bit.
I trialed several data types for output buffer, and
```rust
output: alloc::borrow::Cow<'a, RefCell<Vec<u8>>>,
```

seems to be best. I tried sharing direct reference of `Vec<u8>` and also `Bytes/BytesMut`. 
Direct reference was too difficult to add since encoder uses itself recursively. `BytesMut` was much slower. Maybe this is okay for a while.

Now the buffer is re-used when possible. e.g. no length calculation requirement afterwards.

Current improvement based on the integer test:

image

### DIFF
--- a/src/oer/de.rs
+++ b/src/oer/de.rs
@@ -601,7 +601,6 @@ impl<'input> crate::Decoder for Decoder<'input> {
         _: Tag,
         _: Constraints,
     ) -> Result<Vec<D>, Self::Error> {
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR!

PR #325, opened 2024-09-14
Rework for SetOf type

### LINKED ISSUE #305: `type SetOf<T> = alloc::collections::BTreeSet<T>;` is incorrect
from the [ITU X.680 ASN.1 specification](https://www.itu.int/rec/T-REC-X.680-202102-I/en)

> 3.8.73 set-of types: Types defined by referencing a single component type; each value in the set-of type is an unordered list of zero, one or more values of the component type.

notice that this doesn't mention anything about _deduplication_ of values, only that they are allowed to be encoded in any order, in contrast to `sequence-of` types which mandate they be encoded in the order that they appear in the collection. that means that code like:
```rust
let mut set = SetOf::new();
set.insert(1);
set.insert(1);
```
would be wrong because instead of sending two values, `BTreeSet` would not insert the second value and you'd be left only serializing a single `1`!

Nicceboy:
Fixes #305

Initial idea is to make SetOf as 

```rust
/// The `SET OF` type - an unordered list of zero, one or more values of the component type.
/// Works internally identical to a `Vec<T>`, but with a different name and order of the elements does not matter.
#[derive(Debug, Clone)]
pub struct SetOf<T>(alloc::vec::Vec<T>);
```

It is like SequenceOf, but the all the comparisons ignore the order. It will be costly but I don't know if there is other way around?

### DIFF
--- a/src/ber/de.rs
+++ b/src/ber/de.rs
@@ -24,6 +24,11 @@ type Result<T, E = DecodeError> = core::result::Result<T, E>;
 
 const EOC: &[u8] = &[0, 0];
 
+/// Number of typical re-allocations follows approximately R = ⌈log_g(n/c)⌉ if n > c, else 0
+/// We can assume that in most cases the number of elements is non-zero.
+/// We can reduce re-allocations with larger initial capacity.
+const VEC_BASE_CAPACITY: usize = 4;
+
 /// A BER and variants decoder. Capable of decoding BER, CER, and DER.
 pub struct Decoder<'input> {
     input: &'input [u8],
@@ -419,7 +424,7 @@ impl<'input> crate::Decoder for Decoder<'input> {
         } else if identifier.is_constructed() && self.config.encoding_rules.is_der() {
             Err(DerDecodeErrorKind::ConstructedEncodingNotAllowed.into())
         } else {
-            let mut buffer = Vec::new();
+            let mut buffer = Vec::with_capacity(VEC_BASE_CAPACITY * 2);
 
             if let Some(mut contents) = contents {
                 while !contents.is_empty() {
@@ -468,7 +473,7 @@ impl<'input> crate::Decoder for Decoder<'input> {
         let (input, bs) =
             self::parser::parse_encoded_value(&self.config, self.input, tag, |input, codec| {
                 let Some(unused_bits) = input.first().copied() else {
-                    return Ok(types::BitString::new());
+                    return Ok(types::BitString::with_capacity(VEC_BASE_CAPACITY * 2));
                 };
 
                 match unused_bits {
@@ -604,7 +609,7 @@ impl<'input> crate::Decoder for Decoder<'input> {
         _: Constraints,
     ) -> Result<Vec<D>, Self::Error> {
         self.parse_constructed_contents(tag, true, |decoder| {
-            let mut items = Vec::new();
+            let mut items = Vec::with_capacity(VEC_BASE_CAPACITY);
 
             if decoder.input.is_empty() {
                 return Ok(items);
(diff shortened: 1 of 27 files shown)


Nicceboy:
Actually this is not correct right now. SetOf should not implement `Ord/PartialOrd` I think, as it should be thought as unordered? 

This has impact for some standards which currently derive them and use with the old type.

Nicceboy:
It should be ready if the idea is correct.

XAMPPRocky:
Thank you for your PR! However I think we can simplify the representation and reduce encode work by keeping the Ord requirement. From reading X.680 it only says the order is insignificant, but that doesn't mean that we have to allow any order of elements. Meaning `{1, 2, 3, 4, 5}` equals `{5, 4, 3, 2, 1}`, so to make operations cheap we can require that our internal representation is always ordered, meaning that if you have `SetOf::from([1, 2, 3, 4, 5])` and ``SetOf::from([5, 4, 3, 2, 1])` the internal representation for both is `[1, 2, 3, 4, 5]`.

Nicceboy:
> Thank you for your PR! However I think we can simplify the representation and reduce encode work by keeping the Ord requirement. From reading X.680 it only says the order is insignificant, but that doesn't mean that we have to allow any order of elements. Meaning `{1, 2, 3, 4, 5}` equals `{5, 4, 3, 2, 1}`, so to make operations cheap we can require that our internal representation is always ordered, meaning that if you have `SetOf::from([1, 2, 3, 4, 5])` and `SetOf::from([5, 4, 3, 2, 1])` the internal representation for both is `[1, 2, 3, 4, 5]`.

I thought that at first, but if we allow using push or insert methods, we would need to reorder the set every time afterwards. Or maybe we don't allow that?

Nicceboy:
There is also second problem. SetOf itself should not support `Ord/PartialOrd` because there is no order. However, it makes impossible then to create recursive types, like here without implementing it:

https://github.com/librasn/rasn/blob/90a6603ae7398991442738f79d3f98f28289cc37/standards/ldap/src/lib.rs#L555-L571

Nicceboy:
But if keeping the `Ord` requirement is fine, then the latter is not problem, while semantically it might be a bit misleading.

XAMPPRocky:
> I thought that at first, but if we allow using push or insert methods, we would need to reorder the set every time afterwards. Or maybe we don't allow that?

I think we should use a different collection type other than `Vec` to eliminate this issue, we essentially want `BTreeSet` that allows duplicate values. The simplest version of this would be `BTreeMap<T, usize>` where the value is the number of T's present, but I'm sure there are other more efficient methods.

 > But if keeping the Ord requirement is fine, then the latter is not problem, while semantically it might be a bit misleading.
 
 We need the `Ord` requirement regardless because codecs like DER and PER require that the `SET OF` is sorted in order for the encoding to be canonical.

Nicceboy:
Maybe I could do it with the `HashMap`, it might be more efficient since it does not care about the order like `BTreeMap`. Actually, it does not even require the sorting now that I think of it.

We could implement `to_vec` just by counting the occurences and repeating the same element.

It does not require `Ord/PartialOrd`, so should we drop that implementation?

Something like:

```rust
#[derive(Debug, Clone)]
struct SetOf<T> {
    elements: HashMap<T, usize>,
}

impl<T> SetOf<T>
where
    T: Eq + Hash + Clone,
{
    fn new() -> Self {
        SetOf {
            elements: HashMap::new(),
        }
    }

    fn insert(&mut self, item: T) {
        *self.elements.entry(item).or_insert(0) += 1;
    }

    fn remove(&mut self, item: &T) -> bool {
        if let Some(count) = self.elements.get_mut(item) {
            if *count > 1 {
                *count -= 1;
            } else {
                self.elements.remove(item);
            }
            true
        } else {
            false
        }
    }

    fn contains(&self, item: &T) -> bool {
        self.elements.contains_key(item)
    }
   fn to_vec(&self) -> Vec<T> {
        let mut vec = Vec::with_capacity(self.elements.values().sum());
        for (item, count) in &self.elements {
            for _ in 0..*count {
                vec.push(item.clone());
            }
        }
        vec
    }
}

impl<T> PartialEq for SetOf<T>
where
    T: Eq + Hash,
{
    fn eq(&self, other: &Self) -> bool {
        self.elements == other.elements
    }
}

impl<T> PartialEq<HashMap<T, usize>> for SetOf<T>
where
    T: Eq + Hash,
{
    fn eq(&self, other: &HashMap<T, usize>) -> bool {
        &self.elements == other
    }
}
```

XAMPPRocky:
Yeah, we can go with that.

Nicceboy:
We likely need to use `BTreeMap` after all... since Personnel tests assume that order stays the same, which is not guaranteed by `HashMap` Or those tests should be adjusted quite bit.

XAMPPRocky:
Yeah, the test should probably not rely on the order.

Nicceboy:
I mixed up with sequence of sets and set of sets, there is no need for change. It can be reviewed.

PR #327, opened 2024-09-14
Change LDAPString to be UTF-8

### LINKED ISSUE #304: Should `rasn-ldap` `LdapString` types actually use `String` instead of `Bytes`?
so while reading through the [LDAP RFC](https://datatracker.ietf.org/doc/html/rfc4511#section-4.1.2), I noticed that it mentioned that the `LDAPString` type is encoded as ISO/IEC 10646 which it calls a "superset" of Unicode, however, if we go to the [Unicode FAQ page](https://www.unicode.org/faq/unicode_iso.html) for this, it says:

> Q: What are the differences between ISO/IEC 10646 and Unicode?
> 
> While the character codes and [encoding forms](https://www.unicode.org/glossary/#encoding_form) are synchronized between Unicode and ISO/IEC 10646, the [Unicode Standard](https://www.unicode.org/glossary/#unicode_standard) imposes additional constraints on implementations to ensure that they treat characters uniformly across platforms and applications. To this end, it supplies an extensive set of functional character specifications, character data, [algorithms](https://www.unicode.org/glossary/#algorithm) and substantial background material that is not part of ISO/IEC 10646.

this, to me, says that it would be fine to use `String` instead of whatever backing type for `OCTET STRING` since the code points are defined to be exactly the same, but Unicode _also_ defines algorithms and normalization methods for working with Unicode text, which is orthogonal to the encoding itself, hence the representations, as far as the Rust `String` type is concerned, are equivalent. this would make working with LDAP types much easier, since you would not need to check if the contents are UTF-8 encoded every time you want to get a `&str` or `String` out of them.

Nicceboy:
Fixes #304 

As far as I understand, this was just one line change?

### DIFF
--- a/standards/ldap/src/lib.rs
+++ b/standards/ldap/src/lib.rs
@@ -26,7 +26,7 @@ pub type MessageId = u32;
 /// A notational convenience to indicate that, for `LdapString`, the
 /// ISO10646 character set (a superset of
 /// Unicode) is being used, encoded following the UTF-8 RFC3629 algorithm.
-/// We can use Rust `String` type to represent this type, see
+/// We can use Rust `String` type to represent this type in a form of ASN.1 [`Utf8String`], see
 /// https://github.com/librasn/rasn/issues/304 and https://www.unicode.org/faq/unicode_iso.html
 pub type LdapString = Utf8String;
 
--- a/standards/ldap/src/lib.rs
+++ b/standards/ldap/src/lib.rs
@@ -4,7 +4,6 @@
 extern crate alloc;
 
 use alloc::string::{String, ToString};
-use rasn::error::DecodeError;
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! However I'm not sure. Because this could mean that it is encoded differently by a codec because Utf8String is a different primitive.

I think I would lean towards it being a newtype wrapper around `String` and having a custom encode/decode impl that calls `encode_octet_string`/`decode_octet_string`. Preserving the codec semantics while providing the quality of life improvements.

Nicceboy:
Yeah, tag would have been incorrect for BER, for example. It is changed now.

XAMPPRocky:
Thank you for your PR!

PR #330, opened 2024-09-15
macros: include license texts in the packaged crate

davide125:
The `rasn-derive` crate doesn't currently include the license texts, which is technically a requirement for MIT. This PR ensures they will be included in the packaged crate.


XAMPPRocky:

PR #331, opened 2024-09-19
docs: ObjectIdentifier::new: first arc may be 2

RavuAlHemio:
The first arc is limited to the values 0, 1 and 2, but the doc comment for ObjectIdentifier::new states that an OID is invalid if the first component is greater than 1. (All other instances of such a statement, found mostly around the Oid type, correctly mention 2 as the minimum.)


XAMPPRocky:

PR #338, opened 2024-09-30
Reroll back to BTreeMap

Nicceboy:
Workaround until "final proper" solution for sequence/set field presence tracking is identified.


XAMPPRocky:
Thank you for your PR!

PR #341, opened 2024-10-02
OER: Fix tag decoding and optimise choice/tag encoding further

Nicceboy:
There was a bug when decoding large Tag values, and this fixes that.

Also:

* Removes all allocations related to Tag encoding
* It seems to be possible to use current encoder instance rather than creating new one for Choice encoding which should reduce allocation needs on average.


XAMPPRocky:
Thank you for your PR!

PR #342, opened 2024-10-02
fix: Run clippy and rustdoc only on stable channel in CI

### LINKED ISSUE #334: CI isn't running clips or rustdoc checks
The if statement here doesn't seem working.

https://github.com/librasn/rasn/blob/13a77597156ae57a8f90024bf6696f7abebcea1a/.github/workflows/rust.yml#L91-L94

Nicceboy:
Fixes #334 , at least I hope so.

### DIFF
--- a/src/error/encode.rs
+++ b/src/error/encode.rs
@@ -1,5 +1,5 @@
 //! Error types associated with encoding to ASN.1 codecs.
-#![expect(missing_docs)]
+#![allow(missing_docs)]
 
 use crate::types::constraints::{Bounded, Size};
 use num_bigint::BigInt;
--- a/src/error/encode.rs
+++ b/src/error/encode.rs
@@ -1,6 +1,4 @@
 //! Error types associated with encoding to ASN.1 codecs.
-#![allow(missing_docs)]
-
 use crate::types::constraints::{Bounded, Size};
 use num_bigint::BigInt;
 use snafu::Snafu;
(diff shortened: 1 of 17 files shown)


Nicceboy:
The root level `--workspace` parameter does not catch all the warnings in standards, and I have to check each of them separately. I am not sure if this is a bug or if I am missing something.

Nicceboy:
I could not find a PR or issue, but this was not problem for with the beta version anymore.

src/error/encode.rs
```diff
@@ -1,5 +1,5 @@
 //! Error types associated with encoding to ASN.1 codecs.
-#![expect(missing_docs)]
+#![allow(missing_docs)]
```
XAMPPRocky, line 2: This should be expect because it should be removed when it is no longer true.

Nicceboy: It fails the pipeline alone with:

```
error: this lint expectation is unfulfilled
 --> src/error/encode.rs:2:11
  |
2 | #![expect(missing_docs)]
  |           ^^^^^^^^^^^^
  |
  = note: `-D unfulfilled-lint-expectations` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(unfulfilled_lint_expectations)]
  ```
  
  Should we add `#[allow(unfulfilled_lint_expectations)]` as well to the module?

Nicceboy: I improved the error documentation a bit so that it can be removed. I also removed couple errors since they were not used anymore. There are likely more. Documentation improvements are rather minimal and could be improved in the future.

PR #348, opened 2024-10-19
chore: fix formatting errors in cbr

kevinji:
This fixes the build errors.


XAMPPRocky:
Thank you for your PR! However I already fixed it on main before seeing your PR, thank you for the quick fix though!

PR #352, opened 2024-10-23
Disallow structs without fields for a `set`

shahn:
Also fix an oversight from the syn2 migration that wasn't caught in the testcases.


XAMPPRocky:
Thank you for your PR!

Issue #193, opened 2023-11-02
Traits for macro expansion not always correctly carried with generics when using Encode/Decode


Nicceboy:
It seems that syntax matters for declaring generics and trait declarations.

I have the following struct:

```rust
#[derive(AsnType, Encode, Decode, Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
#[rasn(choice, automatic_tags)]
pub enum RPermissions<T: CertExtType> {
    Content(T::Req),
    All(()),
}
```

It does not compile, and if I expand `Encode` macro for example, `CertExtType` condition is not carried.

However, if I change syntax to following

```rust
#[derive(AsnType, Encode, Decode, Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
#[rasn(choice, automatic_tags)]
pub enum RPermissions<T>
where
    T: CertExtType,
{
    Content(T::Req),
    All(()),
}
```

Correct macro code is generated. I almost ended up writing custom functions, but noticed by accident. If someone else fights the same problem 😁 .

XAMPPRocky:
Huh, the first code should work.

Nicceboy:
> Huh, the first code should work.

It seems to work when newtype style is used, for example

```rust
#[derive(AsnType, Debug, Decode, Encode, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
#[rasn(automatic_tags)]
pub struct ContributedExtensionBlocks<T: ContributedExtension>(
    SequenceOf<ContributedExtensionBlock<T>>,
);
```

XAMPPRocky:
@Nicceboy Is this still the case?

Nicceboy:
Yeah, the commented in the following does not compile:

```rust
﻿use rasn::prelude::*;

pub trait TestTrait: Encode + Decode + AsnType {
    type Req: Encode + Decode + AsnType;
}

#[derive(AsnType, Encode, Decode, Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
pub struct Hello(String);

// #[derive(AsnType, Encode, Decode, Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
// #[rasn(choice, automatic_tags)]
// pub enum TraitNotCarried<T: TestTrait> {
//     Maybe(T::Req),
//     MaybeNot(()),
// }

[...]

PR #354, opened 2024-10-25
fix: Don't override generics impl in proc-macros on encode impl

Nicceboy:
Fixes https://github.com/librasn/rasn/issues/193


Nicceboy:
Third party crate fails the build currently.

XAMPPRocky:
I'm going to merge this and we can address that separately, maybe we can exclude bench and bench related reps from some platforms.

PR #356, opened 2024-10-28
fix: remove version pin from release-plz

Nicceboy:
Maybe the latest version works now. (https://github.com/MarcoIeni/release-plz/issues/1791#issuecomment-2440815911)


XAMPPRocky:
Thank you for your PR!

PR #360, opened 2024-10-28
bench: use black box on integer test, change for macros

Nicceboy:
Wraps the integer bench cases with `black_box` method to possibly prevent some compiler optimizations which are not realistic.


XAMPPRocky:
Thank you for your PR!

PR #362, opened 2024-10-29
fix: cleanup some unnecessary dependencies

### LINKED ISSUE #359: rasn needlessly pulls in rayon as a dependency
When I add a crate, I normally try to minimize the number of new dependencies that get pulled in.

When doing so, I noticed that rasn-derive depends on rayon. Investigating further shows that it only uses rayon once, to do a more convenient flat_map-operation. See: https://github.com/librasn/rasn/blob/main/macros/src/config.rs#L1259

Is this really necessary? Would a PR that removes this dependency be welcomed? I can do this maintenance task for you and change the offending passage to only depend on std.

Also, does rasn 0.18.0 actually not compile when configuring "default-features = false"? I'll need the derives, so it doesn't really affect me, but I noticed this because I normally start with no default features and then add what I need as I go.

Nicceboy:
Fixes #359 

Seems like `rayon` is only used for processing constraints in parallel and typically the amount of constraints is rather small.
I could not measure any drastic performance difference when compiling whole workspace, so I just removed `rayon`:

```
# With rayon
cargo build --workspace --all-targets  142.49s user 17.37s system 742% cpu 21.517 total

# Without rayon
cargo build --workspace --all-targets  141.91s user 17.11s system 752% cpu 21.136 total
```

Also `arrayvec` has been removed as I could not find any usage for it.

There is constant equivalent for current `konst` dependency usage, so this removes that as well.


Nicceboy:
I also removed the `macros` feature since it is not currently possible to compile `rasn` without it. Or should we rework `Open` type before doing so?

XAMPPRocky:
Thank you for your PR!

PR #363, opened 2024-10-29
Remove dependency on rayon

### LINKED ISSUE #359: rasn needlessly pulls in rayon as a dependency

decryphe:
Fixes #359

As discussed in that issue, removes rayon from the dependencies, making the crate a little more lightweight to install.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -312,7 +312,6 @@ dependencies = [
  "once_cell",
  "oorandom",
  "plotters",
- "rayon",
  "regex",
  "serde",
  "serde_derive",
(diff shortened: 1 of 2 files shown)


decryphe:
This should go pretty well together with PR #362 which partially removes rayon as a dependency as well (but doesn't in the dev-dependencies).

XAMPPRocky:
Thank you for your PR! Can you remove the changelog change? That's autogenerated by the release tool.

decryphe:
Rebased and pushed

XAMPPRocky:

decryphe:
Cool, thanks a lot!

PR #364, opened 2024-10-30
Extract proc macro impl into own crate

shahn:
This helps testability of the proc macro, allows fuzzing it to find panics, and would also allow benchmarking.

This might also help investigations like in #362 and #363 to see if there are actual performance changes to the derive macros.

Careful review is probably necessary, because a new crate is introduced. I am happy to take this into any direction, but it is also the foundation for the error message improvements for the derive macros I am working on.


shahn:
I guess another idea would be to do it like serde does it, using a non-published crate that has a symbolic link for the actual implementation, to use the non-published crate for tests etc and publish the rest like normal. I guess that would be preferable, here, too. I will follow up with a commit to add that

shahn:
Hrm, that has the downside of Windows (and serde actually publishes the serde_derive_internals crate, too). Do you have any preference?

XAMPPRocky:
We can just go with a separate crate for now.

PR #365, opened 2024-10-30
Attempt to fix base benchmarks on CI

shahn:
This should install valgrind, which iai and iai-callgrind need to run.


XAMPPRocky:
Thank you for your PR!

shahn:
Unfortunately that wasn't enough as it seems no bencher project was set up (or at least isn't correctly mentioned in https://github.com/librasn/rasn/blob/0d4858dde28f9cc24d63f1720d25de8dded90b99/.github/workflows/base_benchmarks.yml#L19)

Nicceboy:
I think that current issue is on mixing up different kind of benches (criterion/iai-callgrind) at the same time so it does not know how to parse them correctly.

PR #366, opened 2024-10-30
fix: split iai and criterion benches, use correct adapters

Nicceboy:
One attempt more to see if benches work. After there is some historical data, we can change this to apply all PRs?

### DIFF
--- a/.github/workflows/base_benchmarks.yml
+++ b/.github/workflows/base_benchmarks.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: bencherdev/bencher@main
       - run: cargo install iai-callgrind-runner@0.14.0
       - run: sudo apt update && sudo apt install -y valgrind
-      - name: Track base branch benchmarks with Bencher
+      - name: Track base branch IAI benchmarks with Bencher
         run: |
           bencher run \
           --project rasn \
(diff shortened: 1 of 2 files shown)


shahn:
Just a question, why did you set bench = false for the iai-bench?

Nicceboy:
> Just a question, why did you set bench = false for the iai-bench?

Running `cargo bench` will collect all the benches with `bench = true`, so I wanted to avoid maintenance burden to explicitly call the criterion-based tests, and `iai-bench` is called alone explicitly. Maybe there is a better way, however.

shahn:
Thank you for the explanation! IMO, for the non-CI-user, it seems a bit surprising that some of the benches aren't run when using cargo bench - perhaps the maintenance burden is not that great to explicitly decide what to run in CI? But that is not for me to decide :)

Nicceboy:
With the current state of the codecs, I haven't just found IAI bench very useful yet since there are so many improvements left on the logic level and allocations, before we can get benefits out from the IAI by going deeper, so I was kinda not prioritizing it. But I can change it to remain by default on.

Nicceboy:
Valgrind support on MacOS and Apple Silicon is coming closer (https://github.com/LouisBrunner/valgrind-macos/issues/56) so it should be also easier to use soon

shahn:
Ah, I guess disabling it until valgrind is available for those on Apple Silicon is something I didn't consider above

epompeii:
Hey all, Bencher maintainer here. I'm glad things are working!
In the future, feel free to @epompeii me if you ever have any questions or feedback.

PR #370, opened 2024-11-22
Add lifetime for `encoder` trait and add allocation improvements based on that (OER)

Nicceboy:
If we add lifetime for `Encoder` trait, we can recursively share the same output buffer for new encoders with the help of `AnyEncoder` associated type, without using `RefCell` and `Rc` combination.

We can avoid the runtime performance of `RefCell` and can guarantee that mutable reborrow does not happen in compile time. 

Currently only OER does this, but it applies for other codecs too, if they will be also optimized. 

I was also planning on adding `encode_buf` method for OER to reuse the same buffer, since it is trivial after this change, and maybe we can add something like that for other codecs too in small steps.


XAMPPRocky:
> Currently only OER does this, but it applies for other codecs too, if they will be also optimized.

Can you make tracking issues for this and other optimisations that have so far only been implemented for OER? I feel like the other codecs have a lot of catch up to do.

Nicceboy:
> > Currently only OER does this, but it applies for other codecs too, if they will be also optimized.
> 
> Can you make tracking issues for this and other optimisations that have so far only been implemented for OER? I feel like the other codecs have a lot of catch up to do.

I can add some notes later. Currently the primary focus have been just on attempting to reduce the allocations for theoretical minimum.

Nicceboy:
This will be likely the last performance optimization for OER in a while, after this PR it is around 20-30% faster than asn1c.

Nicceboy:
I think this can be reviewed as well.
There are no easy optimizations left anymore. Those constraints could be improved further, but other than that there is branch prediction and unsafe left. Or maybe see if some in-house parser is faster than `nom` since in OER the use-case is very simple, and check the penalty of some `BitView/BitSlice` operations. 

In most cases encoding operations uses only two buffers now, regardless of the type complexity, and all the dynamic data is located in there (excluding BigInt temporaly). 
Decoding allocates only for types that require it. 

There is `encode_buf` function for OER now. 

image

Nicceboy:
When looking back to the dudycz's non-extended integer test, encoding speed is now around 25 faster and decoding speed around 50 times faster

image

Nicceboy:
I still removed some BitSlice usages (preamble & extension header) since the difference was measurable:

image

PR #375, opened 2024-11-23
OER: improve decoding field presence tracking

Nicceboy:
Similar than #324 but for decoding, and OER only for now.

Also fixes casing in one pro-macro variable.

Extended integer test:

image


XAMPPRocky:
Thank you for your PR! 🎉

PR #384, opened 2024-11-28
OER: remove nom usage, improve errors

Nicceboy:
OER decoding has simplified a bit and we can actually just remove `nom` usage completely.
This improves error messages slightly as well at the same time, and move some historical constants.


XAMPPRocky:
Thank you for your PR!

PR #394, opened 2024-12-30
docs: fix accidental double backtick in `Integer` docs

repnop:
noticed the broken formatting while browsing the docs, this also adds a doc link to `isize` for convenience.


XAMPPRocky:
Thank you for your PR!

PR #398, opened 2025-01-03
fix(derive): use `const` blocks and add generic bounds when generating an enum's `AsnType` impl

### LINKED ISSUE #397: `AsnType` derive fails to compile for generic enums
reproduction on the current (0.22.0) version:
```rust
use rasn::{types::OctetString, AsnType};

#[derive(AsnType)]
#[rasn(choice)]
pub enum Example<T> {
    String(OctetString),
    Sequence(Vec<T>),
}
```
produces the errors
```
error[E0401]: can't use generic parameters from outer item
 --> src/lib.rs:7:18
  |
5 | pub enum Example<T> {
  |                  - type parameter from outer item
6 |     String(OctetString),
7 |     Sequence(Vec<T>),
  |                  ^ use of generic parameter from outer item
  |

error[E0277]: the trait bound `T: AsnType` is not satisfied
 --> src/lib.rs:7:14
  |
7 |     Sequence(Vec<T>),
  |              ^^^^^^ the trait `AsnType` is not implemented for `T`, which is required by `Vec<T>: AsnType`
  |
  = note: required for `Vec<T>` to implement `AsnType`
help: consider restricting type parameter `T`
  |
5 | pub enum Example<T: rasn::AsnType> {
  |                   +++++++++++++++
```

looks like `TAG_TREE` needs to use a `const` block which will allow it to access the generic parameter, and also needs to add a bound for `T` by default to bound it for `AsnType`. I'll work up a PR to fix these.

repnop:
fixes #397. this commit ensures that the `TAG_TREE` associated const uses a `const` block so it is able to access any outer generics on the enum and also bounds any type generics with `AsnType` so that impls will be met.

### DIFF
--- a/macros/macros_impl/src/asn_type.rs
+++ b/macros/macros_impl/src/asn_type.rs
@@ -62,7 +62,7 @@ pub fn derive_struct_impl(
                 paren_token: Some(<_>::default()),
                 modifier: syn::TraitBoundModifier::None,
                 lifetimes: None,
-                path: syn::parse_str("rasn::AsnType").unwrap(),
+                path: syn::parse_quote!(#crate_root::AsnType),
             }));
     }
 
--- a/macros/macros_impl/src/enum.rs
+++ b/macros/macros_impl/src/enum.rs
@@ -39,10 +39,10 @@ impl Enum {
 
             quote! {
                 {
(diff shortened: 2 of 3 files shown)


XAMPPRocky:
Thank you for your PR!

PR #400, opened 2025-01-04
Better error messages for derives

shahn:
I added error span reporting to the derive macros. While doing so, I was worried about degrading performance, so I added a benchmark. To my surprise, performance improved between 2%  and 20% for each of my individual benches. Benches are auto-generated from the rasn code base, by simply parsing all structs and enums which are using derives.

I also added a fuzz test for the derives, to help see if I missed any panics which would definitely not lead to good errors.

Please see which of the commits in this PR you want to take, because only the last one is about the better error messages. I am also happy to provide individual PRs if you think this would make sense. It is probably not useful to just squash all these into a single merge commit, as that removes the ability to reproduce the benchmark.


XAMPPRocky:
Thank you for your PR!

XAMPPRocky (approved):

shahn:
I am still working on this, there are some remaining instances of bad error messages left which I am weeding out.

shahn:
Should be fine now :)

PR #404, opened 2025-01-08
feat: add a convenience macro for parsing string literals into `&'static Oid`s

repnop:
this is something I had whipped up for one of my own projects a while back, so figured it might be useful to have in `rasn` proper. there is one caveat with the way this is currently implemented, when an error does occur, the error message is pretty brutal (in that it's filled with a lot of useless `note`s, but I made an issue on the Rust repo about this, so it should get fixed at some point, as it looks like a regression even: https://github.com/rust-lang/rust/issues/135259).

let me know if there's any adjustments or fixes desired for the impl.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -186,3 +186,80 @@ macro_rules! bounded_constraint {
         )
     };
 }
+
+/// Helper macro to create a `&'static Oid` from a string literal. The string
+/// literal must conform to the standard OID format and must be a valid OID
+/// (first arc must be <= 2), but also accepts OIDs with and without a leading
+/// `.`.
+///
+/// Usage:
+/// ```rust
+/// const SYS_DESCR: &'static rasn::types::Oid = rasn::oid!(".1.3.6.1.2.1.1.1.0");
+/// assert_eq!(SYS_DESCR, rasn::types::Oid::new(&[1, 3, 6, 1, 2, 1, 1, 1, 0]).unwrap());
+/// ```
+#[macro_export]
+macro_rules! oid {
+    ($s:literal) => {{
+        const OID: &'static $crate::types::Oid = const {
+            const BYTE_STRING: &'static [u8] = const {
+                let s: &'static str = $s;
+                let bytes = s.as_bytes();
+
+                core::assert!(!bytes.is_empty(), "OID string literals cannot be empty");
+                core::assert!(bytes.is_ascii(), "OID string literals must be ASCII");
+                core::assert!(bytes[bytes.len() - 1] != b'.', "OID string literals cannot end with a period");
+
+                match bytes {
+                    [b'.', rest @ ..] => rest,
+                    bytes => bytes,
+                }
+            };
+
+            const COMPONENT_LEN: usize = const {
+                let mut component_count = 1;
+                let mut index = 0;
+                while index < BYTE_STRING.len() {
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! I think my initial question, is that, is there a reason it's a macro as opposed to a `const fn`? It seems like it doesn't need to be a macro.

repnop:
so there's few problems fighting against this being a `const fn` instead:

1. there is no way to dynamically size the output slice based on a function parameter without `#![feature(unsized_locals)]` (which would also need to be available in `const fn` when stabilized)
2. there's no way to produce a `&'static [u32]` without `const`s inside of the function regardless, but this isn't possible because you can't use a parameter as input to a `const` item since it's possible for the value to be runtime determined
3. `&'static str`s are not a valid `const` generic parameter (which prevents passing the value as `Oid::const_parse::<"1.2.3">()`) and even if they were, I don't believe it is possible to reference the `const` generic parameter from `const`s inside of the function, since they're supposed to be independent of any input parameters and you'd need to create a local `let` binding to use them (which then cannot be promoted to a `&'static` value)

hopefully this snippet is illustrative of the problems:
```rust
const fn parse_str(s: &'static str) -> &'static [u32] {
    // error[E0435]: attempt to use a non-constant value in a constant
    //  --> src/main.rs:2:38
    //   |
    // 1 | const fn parse_str(s: &'static str) -> &'static [u32] {
    //   |                    - this would need to be a `const`
    // 2 |     let mut array_for_slice = [0u32; s.len()];
    //   |                                      ^
    let mut array_for_slice = [0u32; s.len()];
    
    // if the above was possible, how to promote `array_for_slice` to a
    // `&'static`?
}

// error[E0401]: can't use generic parameters from outer item
//   --> src/main.rs:17:30
//    |
// 15 | const fn parse_str2<const S: &'static str>() -> &'static [u32] {
//    |                           - const parameter from outer item
// 16 |     const ARRAY: &[u32] = const {
// 17 |         let mut arr = [0u32; S.len()];
//    |                              ^ use of generic parameter from outer item
//    |
//    = note: a `const` is a separate item from the item that contains it
// 
// error: `&'static str` is forbidden as the type of a const generic parameter
//   --> src/main.rs:15:30
//    |
// 15 | const fn parse_str2<const S: &'static str>() -> &'static [u32] {
//    |                              ^^^^^^^^^^^^
//    |
//    = note: the only supported types are integers, `bool`, and `char`
const fn parse_str2<const S: &'static str>() -> &'static [u32] {
    const ARRAY: &[u32] = const {
        let mut arr = [0u32; S.len()];
        // do parsing work here
        arr
    };
    
    ARRAY
}
```

XAMPPRocky:
> let mut array_for_slice = [0u32; s.len()];

You could get around this by just setting ` let mut array_for_slice = [0u32; 256];`, we don't need to scale an array dynamically, we just need to have enough space, and if someone has an OID that is over 256 components long we can double it 😄

repnop:
I agree, however that does come with the problem of how to actually promote the local value to a `&'static static` so it can be used for literal values, which isn't possible without allocating _and_ leaking the allocated value :( I was going to mention doing that but ended up not because it doesn't really get much further along.

XAMPPRocky:
We can keep it as is then for now. Thank you for your PR!

PR #405, opened 2025-01-08
fix: some clippy lints causing CI to fail

repnop:
applied clippy's suggestions for the lints, hopefully fixes CI failures


XAMPPRocky:
Thank you for your PR!

PR #406, opened 2025-01-09
Add basic support for REAL using f32/f64 with JER/OER rules

kodprussiluskan:
I think this is a compliant implementation of REAL, as long as you're fine with being constrained to 32/64-bit floats.

From what I can gather this is the first type that isn't supported for every codec so I added a generic error for the unsupported codecs. Unfortunately this means that you won't get an error until runtime.

I added f32/f64 feature flags so you don't have to use any of it if you don't want it.

Neither my rust nor my asn.1 is something to brag about, so if you want it done in another way just let me know.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,10 @@ chrono = { version = "0.4.38", default-features = false, features = ["alloc"] }
 bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 
 [features]
+default = ["f32", "f64"]
 std = []
+f32 = []
+f64 = []
 backtraces = ["std", "snafu/backtrace"]
 compiler = ["rasn-compiler"]
 
--- a/src/types/real.rs
+++ b/src/types/real.rs
@@ -1,0 +1,81 @@
+use num_traits::float::FloatCore;
+
+/// Represents a real type in Rust that can be decoded or encoded into any
+/// ASN.1 codec.
+pub trait RealType: Sized + core::fmt::Debug + core::fmt::Display {
+    /// The byte level width of the floating point type.
+    const BYTE_WIDTH: usize;
+
+    /// Returns the IEEE 754 encoded bytes of the real type, byte count defined in `usize`
+    fn to_ieee754_bytes(&self) -> (impl AsRef<[u8]>, usize);
+
+    /// Attempts to decode the IEEE 754 encoded bytes into `Self`
+    fn try_from_ieee754_bytes(bytes: &[u8]) -> Result<Self, TryFromRealError>;
+
+    /// Attempts to convert a generic floating point type into `Self`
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -28,7 +28,10 @@ chrono = { version = "0.4.38", default-features = false, features = ["alloc"] }
 bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 
 [features]
+default = ["f32", "f64"]
 std = []
+f32 = []
+f64 = []
```
XAMPPRocky, line 34: Is there a real need for these feature flags? Are there platforms where these aren't available but has alloc support?

src/types/real.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn to_ieee754_bytes(&self) -> (impl AsRef<[u8]>, usize);
+
+    /// Attempts to decode the IEEE 754 encoded bytes into `Self`
+    fn try_from_ieee754_bytes(bytes: &[u8]) -> Result<Self, TryFromRealError>;
+
+    /// Converts the real type into a `f64`.
+    fn to_f64(&self) -> f64;
+
+    /// Converts a `f64` into the real type.
+    fn from_f64(value: f64) -> Self;
```
XAMPPRocky, line 22: These should be `TryInto`/`TryFrom` equivalents meaning that it should allow for the possibility that it fails to convert.

src/types/real.rs
```diff
@@ -0,0 +1,86 @@
+/// Represents a real type in Rust that can be decoded or encoded into any
+/// ASN.1 codec.
+#[cfg(any(feature = "f32", feature = "f64"))]
+pub trait RealType:
```
XAMPPRocky, line 4: Similar to the question above, this trait should always be available even if those types aren't to allow people to use custom float types.

XAMPPRocky (changes requested):
Overall I like this approach for starting out with implementing REAL, since it specifies the ideal interface, I have a couple of comments, but overall LGTM.

Cargo.toml
kodprussiluskan: Embedded targets might have alloc but either have no fp support, or don't want to enable it.

It could be enabled by default anyway though since I'm guessing most people can use it?

src/types/real.rs
kodprussiluskan: See my comment about embedded targets, however, RealType might need some work anyway since you can have f32 support, but no hw support for f64.

XAMPPRocky: Right but you could decode real types into non floating point types. I think we should remove the cfg flags for everything except the trait definitions for f32 and f64

kodprussiluskan: Updated, RealType is now always available, and only the actual implementations are guarded by the f32/f64 feature flags.

PR #410, opened 2025-01-29
feat:add graphic string support

privatepatrick:
Copy of GraphicString removing the control and delete characters


XAMPPRocky:
Thank you for your PR!

Issue #418, opened 2025-03-06
Not handling the encoding of explicit tags when a default value is present


jusbeyer:
Recently rasn added support for explicit tags in an ASN.1 Structure. When this support was added, though, it does not appear that anything was added to handle the non-encoding default values. 

Per ITU standard, this value should not be present during encoding. Since decoding automatically sets the default value, when not present, and will preserve the value, if provided, a fix should be implemented and will be backwards compatible (assuming downstream systems correctly handle this).

Example code:
```rust
use rasn::prelude::*;

#[derive(AsnType, Decode, Debug, Encode, PartialEq)]
pub struct SequenceWithMultiDefault {
    #[rasn(tag(explicit(application, 1)), default="default_bool")]
    b: bool,
    #[rasn(tag(explicit(application, 2)), default="default_bool")]
    b2: bool,
}
pub fn default_bool() -> bool {
    true
}

fn main() {
    let seq = SequenceWithMultiDefault{ b: true, b2: false };
    let encoded = rasn::der::encode(&seq).unwrap();
    println!("{:02x?}", encoded);
}
```

Produces:
```
[30, 0a, 61, 03, 01, 01, ff, 62, 03, 01, 01, 00]
```

Should Produce (Note: The first tagged field is not present):
```
[30, 05, 62, 03, 01, 01, 00]
```

PR #419, opened 2025-03-06
Fixes librasn/rasn#418. Add support for explict tags and default values to encode correctly

jusbeyer:
Currently, when using an explicit tag default values are not considered and only a code path of adding that value to the encoded output is followed.

This change adds support to check if the default value matches and, if so, does not encode it. This will have no impact on the decode path.


jusbeyer:
Looks like PR #416 fixes the clippy issues these tests are failing on. That will need to be merged first and then these tests can be re-run.

XAMPPRocky:
Thank you for your PR! I've now merged #416.

XAMPPRocky:

jusbeyer:
Thank you for the fast response on this!

PR #421, opened 2025-03-08
Add decoding iterator, decode_with_remainder, clippy fixes

### LINKED ISSUE #413: Decode multiple messages from the same buffer
How can I decode multiple UPER-encoded messages from a buffer where they are stored directly after each other, without delimiter? `decode` just takes a buffer, but neither shrinks it when reading bytes nor returning how many bytes it read to parse the first message. Initially I thought the impl of `Decode` for `Vec<T>` where `T: Decode` would help me, but that assumes that the messages are stored in an asn1 sequence rather than immediately one after another. I didn't see any kind of API to read this efficiently, did I overlook that or could it be added like a `decode_multiple` (would require alloc of course) or something like that?

Nicceboy:
This PR adds iterator over decoding bytes (fixes #413) and `decode_with_remainder` public function for binary codecs. 

Iterator works for same types and `decode_with_remainder` is helpful if there is a need to decode different types from the buffer and usage of the iterator is not possible.

Usage is a bit different than in the issue since there was no unified way to construct `Decoder` instances and this approach abstracts that and keeps the instancing in one place.
I guess if we want to use the initial `Decoder` approach, we need to add `new` for the decoder trait and unify all Encoder/Decoder options? 

Current usage, for example:

```rust
let mut iter = crate::de::iter(&data, Codec::Uper);
let decoded: Integer = iter.next().unwrap().unwrap();
```

Iterator container has Borrowed and Owned variants since I tried to keep allocations in zero if no new data is appended.


XAMPPRocky:
Thank you for your PR!

Nicceboy:
PR slipped one coverage-related binary file - need to remove that.

PR #425, opened 2025-03-10
ConstrainedInteger: From -> TryFrom

### LINKED ISSUE #399: `ConstrainedInteger`'s `From` impl allows for creating a `ConstrainedInteger` which is outside of the allowed range
currently the blanket `From<T: Into<Integer>>` impl allows creating an instance of `ConstrainedInteger<START, END>` with a value that exists outside of the constrained value. this should probably be changed to be a `TryFrom` impl instead where the bounds can be verified and `Err` returned if the converted-from value is outside of the range, though this would be a breaking change.

example:
```rust
let bar = rasn::types::ConstrainedInteger::<5, 100>::from(0u32);
```

Nicceboy:
Fixes #399 

Additionally changes inner type of `ConstrainedInteger` to `i128` since it does not currently make much sense to use `Integer` here.

### DIFF
--- a/src/aper.rs
+++ b/src/aper.rs
@@ -215,9 +215,9 @@ mod tests {
         round_trip!(aper, Integer, (-256).into(), &[0x02, 0xff, 0x00]);
         round_trip!(aper, Integer, (-32768).into(), &[0x02, 0x80, 0x00]);
         round_trip!(aper, Integer, (-32769).into(), &[0x03, 0xff, 0x7f, 0xff]);
-        round_trip!(aper, B, 5.into(), &[0x00]);
-        round_trip!(aper, B, 6.into(), &[0x02]);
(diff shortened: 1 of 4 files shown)


XAMPPRocky:
> Additionally changes inner type of ConstrainedInteger to i128 since it does not currently make much sense to use Integer here.

Thank you for your PR! I'm not sure if it does make more sense. I haven't communicated this but I have been experimenting with trying to make `Integer` a single pointer width using pointer provenance, and if we can implement that, then `Integer` will be smaller than i128.

Nicceboy:
Ah, I reverted that change but kept the `isize, i32, i64, i128` only in `try_from` since it forces the type size to be in constraint range and solves the issue with overlapping implementations and is likely enough for all use cases.

PR #428, opened 2025-03-12
macros: Improve constraint compiler errors

Nicceboy:
There was a bug when autogenerating constraints - proc macros added sometimes `i128` suffix for constraint values in delegate context for attributes and these failed the constraint parsing silently in the end. (constraints were not noted at all). This improves invalid value detection in constraint context.

Also added support to express constraint attribute numbers with underscore, like `12_345`.


XAMPPRocky:
Thank you for your PR!

PR #437, opened 2025-03-29
feat: inner subtype constraint

Nicceboy:
Adds `InnerSubtypeConstraint` constraint trait and some relevant error messages. Mainly used in #389 

Usage requires implementing single method which should validate the inner constraints:

```rust
/// Validates the inner subtype constraints and attempts to decode internal ASN.1 `CONTAINING` constraints as a marked type.
/// Usually this means that some field has `OPAQUE` data, and we need to decode it further as a specific type, as defined in the inner subtype constraint.
/// Manual implementation of this function is required for all types that have inner subtype constraints.
/// # Arguments
/// * `decode_containing_with` - the codec to validate and decode the containing data with - ASN.1 type definiton has `CONTAINING.
fn validate_and_decode_containing(
    self,
    decode_containing_with: Option<crate::Codec>,
) -> Result<Self, crate::error::InnerSubtypeConstraintError>;
```

### DIFF
--- a/src/types/constraints.rs
+++ b/src/types/constraints.rs
@@ -3,6 +3,25 @@
 use super::IntegerType;
 use num_bigint::BigInt;
 
+/// A marker trait with validation methods for types that have ASN.1 inner subtype constraints.
+pub trait InnerSubtypeConstraint: Sized {
+    /// Validates the inner subtype constraints and returns the type on success.
+    /// Does not attempt to decode internal ASN.1 `CONTAINING` constraints as a marked type. See `validate_and_decode_containing` for CONTAINING validation.
+    fn validate_components(self) -> Result<Self, crate::error::InnerSubtypeConstraintError> {
+        self.validate_and_decode_containing(None)
+    }
+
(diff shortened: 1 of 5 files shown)


src/types/constraints.rs
```diff
@@ -3,6 +3,25 @@
...
+    /// Does not attempt to decode internal ASN.1 `CONTAINING` constraints as a marked type. See `validate_and_decode_containing` for CONTAINING validation.
+    fn validate_components(self) -> Result<Self, crate::error::InnerSubtypeConstraintError> {
+        self.validate_and_decode_containing(None)
+    }
+
+    /// Validates the inner subtype constraints and attempts to decode internal ASN.1 `CONTAINING` constraints as a marked type.
+    /// Usually this means that some field has `OPAQUE` data, and we need to decode it further as a specific type, as defined in the inner subtype constraint.
+    /// Manual implementation of this function is required for all types that have inner subtype constraints.
+    /// # Arguments
+    /// * `decode_containing_with` - the codec to validate and decode the containing data with - ASN.1 type definiton has `CONTAINING.
```
XAMPPRocky, line 18: Seems like this sentence got cut off.

XAMPPRocky:
Thank you for your PR!

PR #440, opened 2025-04-04
fix: improve constraint performance and bounded intersection

Nicceboy:
We did not cover all the intersection combinations for value/size constraints. This adds them with tests.

I was also able to move constraint intersection in delegate context into different place and now those constraints are completely evaluated in compile-time. Around 15% performance boost at least for IEEE 1609.2 certs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,15 +142,15 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "base64ct"
-version = "1.6.0"
+version = "1.7.3"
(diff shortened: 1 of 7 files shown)


XAMPPRocky (approved):

Nicceboy:
Just a note which we might need to mention somewhere or change in the future; constraints do not support natively the case where the intersection leads into situation where no values are allowed at all.

For value constraint it simply creates bounded range 
```rust
 Bounded::Range {
                start: Some(1),
                end: Some(-1),
            },
```

And for size constraint the size is set to be 0.

XAMPPRocky:
I feel like we should make an issue for it and we should support that case.

Nicceboy:
Added #442

PR #441, opened 2025-04-05
Replace CI scripts with justfile which can be used locally as well

Nicceboy:
This should make local development more consistent and maybe easier for new contributions as well.

E.g. to run `fmt` and `clippy` with correct parameters, you can just run `make check`.

To run all tests that also CI does, run `make all`.


XAMPPRocky:
Thank you for your PR! I'm very in favour of the idea, but I have a personal aversion to Makefiles, I don't like maintaining and using make. Would you consider moving the PR to just instead? That is a much more modern alternative that is popular in the community written in Rust, so it's easier to contribute fixes and report bugs.

Nicceboy:
I guess I can move to `just` if you want. I took Make since it is available probably in every system and it is so old that there are likely never issues with the runner itself, if we need to consider the bugs. On average most are likely more familiar with that as well.

XAMPPRocky:
I think that would make sense if Rust wasn't already a requirement to contribute, so if you're already installing Rust, it's not a big hurdle to ask to install a tool that can be installed with Rust through cargo.

Nicceboy:
Can't say that I am fan yet, but there is `justfile` now.

XAMPPRocky:
Thank you for your PR!

PR #443, opened 2025-04-06
Change from TagTree in choice for Tag array

Nicceboy:
I am still doing some benchmarking and optimizations, and it seems that `TagTree::tag_contains` takes a lot time when there are many `Choice` types in the overall type.

It seems to call itself recursively at least once no matter what it contains.
This adds ups since when especially decoding `Choice` variant that is the last in order, `tag_contains` can be called 1 + amount of variants or even more.  

So I have tried to replace `TagTree` with simple `&'static [Tag]` whenever possible, including in Choice trait, and there is a pretty significant performance boost available.

I still have to test if there are some side effects what current tests are not showing. 

image

At least current codec implementations just want to know whether tag exists with VARIANTS/EXTENDED_VARIANTS so maybe this can be changed.

### DIFF
--- a/macros/macros_impl/src/config.rs
+++ b/macros/macros_impl/src/config.rs
@@ -653,9 +653,16 @@ impl<'config> VariantConfig<'config> {
                 })
             }
         };
+        let (tag_tokens, is_tag_tree) = tag_tree;
+        let if_check = if is_tag_tree {
+            quote!(#crate_root::types::TagTree::tag_contains(&tag, &[#tag_tokens]))
+        } else {
+            quote!(#tag_tokens.eq(&tag))
+        };
 
         Ok(quote! {
-            if #crate_root::types::TagTree::tag_contains(&tag, &[#tag_tree]) {
+
+            if #if_check {
                 #const_constraint
                 return #decode_op
             }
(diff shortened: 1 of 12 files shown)


XAMPPRocky:
>At least current codec implementations just want to know whether tag exists with VARIANTS/EXTENDED_VARIANTS so maybe this can be changed.

Doesn't this affect SET implementations though since they need to be able figure out what the tag is for each component and there could be multiple nested choice types?

Nicceboy:
For nested choices it currently uses conditionally `TagTree::tag_contains` call, which should be enough for that, and reduces the overhead from everywhere else. This happens on derived code rather than on codec level. But that needs more tests I think.

Nicceboy:
Also note - if `&'static [Tag]` is valid and can be used, in many cases tags can also be sorted and we could use `binary_search` instead. On few variants there is no speedup but if someone uses extremely many variants, it would be visible. Premature optimization?

Nicceboy:
This still on the list, I just have been a bit busy. Maybe we can do binary search part later.

Nicceboy:
I have finally time to work with this properly again.

PR #449, opened 2025-04-15
Add CMS signed attribute CMSAlgorithmProtection.

mt-johan:
[RFC 6211](https://www.rfc-editor.org/rfc/rfc6211#section-2):

```text
   aa-cmsAlgorithmProtection ATTRIBUTE ::= {
       TYPE CMSAlgorithmProtection
       IDENTIFIED BY { id-aa-CMSAlgorithmProtection }
   }

   id-aa-CMSAlgorithmProtection OBJECT IDENTIFIER ::= { iso(1)
        member-body(2) us(840) rsadsi(113549) pkcs(1) pkcs9(9) 52 }

   CMSAlgorithmProtection ::= SEQUENCE {
       digestAlgorithm         DigestAlgorithmIdentifier,
       signatureAlgorithm  [1] SignatureAlgorithmIdentifier OPTIONAL,
       macAlgorithm        [2] MessageAuthenticationCodeAlgorithm
                                        OPTIONAL
   }
   (WITH COMPONENTS { signatureAlgorithm PRESENT,
                      macAlgorithm ABSENT } |
    WITH COMPONENTS { signatureAlgorithm ABSENT,
                      macAlgorithm PRESENT })
```

Note to reviewer: Since `signatureAlgorithm` and `macAlgorithm` are mutually exclusive there might be a better way to model this.


Nicceboy:
There is a new marker trait which allows implementing validator for construction/decoding for such cases, maybe it could help. For this scenario parameter `decode_containing_with` could be likely ignored. 

https://github.com/librasn/rasn/blob/66ef46e6efa4392d089c5247c3604064b2677658/src/types/constraints.rs#L7-L23

Nicceboy:
It was just recently added so in the main branch there aren't any.

There are many examples in the #389 

Sample ASN.1 where at least one must be present

```asn1
SignedDataPayload ::= SEQUENCE { 
  data        Ieee1609Dot2Data OPTIONAL,
  extDataHash HashedData OPTIONAL,
  ...,
  omitted     NULL OPTIONAL
} (WITH COMPONENTS {..., data PRESENT} |
   WITH COMPONENTS {..., extDataHash PRESENT} |
   WITH COMPONENTS {..., omitted PRESENT})

```

https://github.com/librasn/rasn/blob/f513a2110389c9766c3f98d363835224ddb88479/standards/its/src/ieee1609dot2/mod.rs#L173-L217

I think in your case you want to restrict the construction with `new()` method and that makes the validation check. Decoding validation is left for the user so they can optionally call the `validate_components()` for the instanced type.

XAMPPRocky:
Thank you for your PR! And thank you @Nicceboy for the review.

PR #455, opened 2025-04-27
Fix default value of TrustAnchorInfoVersion

### LINKED ISSUE #454: Default value of TrustAnchorInfo.version does not align with RFC 5914 definition
The ASN.1 definition from [RFC 5914](https://datatracker.ietf.org/doc/html/rfc5914#section-2) defines the default value of `TrustAnchorInfo.version` as `v1` of `TrustAnchorInfoVersion`, which has a integer value of 1. 

However, the definition in the `pkix` crate does not implement Default for `TrustAnchorInfoVersion`, so the default value is 0.

I'm happy to put together a PR that:
1. Defines `TrustAnchorInfoVersion` as a newtype of `Integer`
2. Create a `TrustAnchorInfoVersion` impl that defines V1 and `raw_value` in the same manner as `Version`
3. Implement `Default`
4. Create a simple test that exercises the above changes for correctness

Please let know if this approach is good or if a different approach should be taken. Thanks!

turbolight-solutions:
Resolves #454.


XAMPPRocky (approved):

CBonnell:
@XAMPPRocky I noticed that the BER/DER encoder is emitting the following octets for an empty BIT STRING:
`0x03 0x00`.

According to X.690, clause 8.6.2.3, the initial octet still needs to be included and set to 0:

> 8.6.2.3 If the bitstring is empty, there shall be no subsequent octets, and the initial octet shall be zero.

Thus, I believe the correct encoding is `0x03 0x01 0x00`. I confirmed that `pyasn1` produces this encoding.

Obviously this is a separate issue from this PR, but I want to get your thoughts on how to proceed (file a separate issue or if you believe the current behavior is correct, etc.)

XAMPPRocky:
Seems like a bug, would accept an additional pr fixing it.

XAMPPRocky:

turbolight-solutions:
Thanks, I just filed #457 for the empty BIT STRING issue.

PR #458, opened 2025-04-27
Always emit and accept initial octet for empty BIT STRING

### LINKED ISSUE #457: BER encoder and decoder does not emit or accept the initial octet of empty BIT STRINGs
The BER encoder is emitting the following octets for an empty BIT STRING: `0x03 0x00`.

According to X.690, clause 8.6.2.3, the initial octet still needs to be included and set to 0:

> 8.6.2.3 If the bitstring is empty, there shall be no subsequent octets, and the initial octet shall be zero.

Thus, the correct encoding is `0x03 0x01 0x00`. I confirmed that pyasn1 produces this encoding.

Likewise, the BER decoder needs to accept empty BIT STRINGs with the initial octet set to "0".

turbolight-solutions:
Resolves #457.


XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR!

PR #451, opened 2025-04-16
RFC 3161 Time-Stamp Protocol (TSP) ASN.1 module. 

mt-johan:
Contributed under SPDX "MIT OR Apache-2.0" licenses.

See also:

* [RFC 3161](https://www.rfc-editor.org/rfc/rfc3161) Internet X.509 Public Key Infrastructure Time-Stamp Protocol (TSP)
* https://github.com/orgs/librasn/discussions/450

PR #461, opened 2025-05-07
Fix invalid mapping of PKIStatus in TSP. 

mt-johan:
Contributed under SPDX "MIT OR Apache-2.0" licenses.

Mapping PKIStatus to an enum didn't work out as expected in https://github.com/librasn/rasn/pull/451 when doing interoperability tests:

```text
"Error while decoding data. kind: 'FieldError { name: \"TimeStampResp.status\", nested: DecodeError { kind: FieldError { name: \"PkiStatusInfo.status\", nested: DecodeError { kind: CodecSpecific { inner: Ber(MismatchedTag { expected: Tag { class: Universal, value: 10 }, actual: Tag { class: Universal, value: 2 } }) }, codec: Ber } }, codec: Der } }', codec: 'Der'"
```

Revert to using the one-to-one mapping from [RFC 3161 2.4.2](https://www.rfc-editor.org/rfc/rfc3161#section-2.4.2) :

```text
   PKIStatus ::= INTEGER {
      ...
```

### DIFF
--- a/standards/cms/src/tsp.rs
+++ b/standards/cms/src/tsp.rs
@@ -122,20 +122,7 @@ pub struct PkiStatusInfo {
        -- notification that a revocation has occurred  }
 ```
 */
-#[derive(AsnType, Clone, Copy, Debug, Decode, Encode, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[rasn(enumerated)]
-pub enum PkiStatus {
-    /// A TimeStampToken, as requested, is present.
-    Granted = 0,
-    /// A TimeStampToken, with modifications, is present.
-    GrantedWithMods = 1,
-    Rejection = 2,
-    Waiting = 3,
-    /// This message contains a warning that a revocation is imminent.
-    RevocationWarning = 4,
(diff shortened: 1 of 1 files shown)


Nicceboy:
I took a look for the RFC and there could be also alternative way (maybe less explicit) to do this if we add `const_new` function for the `Integer` type when value fits for `isize`. 

It would be possible to make `newtype` of `Integer`, where the `impl` section just defines those values as `const`, and the instancing of the type would be only possible with these constants. Encoding and decoding would be always correct. I am not sure if this would be clearer approach, but the `enum` way maybe be confusing with actual `enum` types .

XAMPPRocky:
I think this problem is worth solving as it is common that there is a functionally limited set of values encoded in a non enumerated type.

I don't have the answer of what the api should be, but as long as it respects the underlying type (e.g. if there can be more values, then it needs to be non exhaustive.) Then I'm all in favour of it.

XAMPPRocky:
> Are you ok with spawning a new issue for this?

Sure go ahead. I'll merge this as-is for now. Thank you for your PR!

PR #444, opened 2025-04-07
chore: clippy pendantic fixes

Nicceboy:
Fixes some clear clippy pedantic suggestions but not all of them.

PR #466, opened 2025-05-22
fix(uper): empty outer encoding

6d7a:
Returns a single zero-octet for empty outermost encodings.

11.1.3.1	(The result of the encoding is not contained in an ASN.1 bitstring) If the result of encoding the outermost value is an empty bit string, the bit string shall be replaced with a single octet with all bits set to 0. If it is a non-empty bit string and it is not a multiple of eight bits, (one to seven) zero bits shall be appended to it to produce a multiple of eight bits.

Technically, there's also the following paragraph to consider, but given that we `rasn::uper::encode` always returns a byte vector that may be padded it does not apply.

11.1.3.2	(The result of the encoding is contained in an ASN.1 bitstring) If the result of encoding the outermost value is an empty bit string, the bit string shall be replaced with a single bit set to 0. No padding bits shall be appended.

### DIFF
--- a/src/uper.rs
+++ b/src/uper.rs
@@ -27,7 +27,12 @@ pub fn decode_with_remainder<T: crate::Decode>(
 pub fn encode<T: crate::Encode>(
     value: &T,
 ) -> Result<alloc::vec::Vec<u8>, crate::error::EncodeError> {
-    crate::per::encode(enc::EncoderOptions::unaligned(), value)
+    let result = crate::per::encode(enc::EncoderOptions::unaligned(), value)?;
+    if result.is_empty() {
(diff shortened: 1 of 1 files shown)


XAMPPRocky (approved):

6d7a:
My bad. I re-rebased now, so all changes from `main` should be on this branch now. Still the actions fail. Do you have an idea what I might be missing?

Nicceboy:
CI issues are fixed in https://github.com/librasn/rasn/pull/444 if it gets merged. (Warnings about dead code)

XAMPPRocky:
Thank you for your PR!

PR #469, opened 2025-06-10
refactor(octet-string): switch from bytes to arc-slice

XAMPPRocky:
Might be of interest to @Nicceboy Locally I got a 10% performance increase in OER and BER.


Nicceboy:
Nice! I can confirm around 7% increase with LTO and 7-10% with default release mode. UPER decode seems to get regression of ~7%.

I wonder is this is worth of some new dependencies and crate seems to be sill relatively new? Maintenance may be concern if the is a lot of unsafe code.

XAMPPRocky:
I updated the code so that the implementation is no longer public and so that the implementation can now be toggled with a feature flag with the default remaining bytes for now.

PR #477, opened 2025-06-17
reexport rasn default-features in standard crates

wyfo:
It makes notably possible to use `arc-slice` feature with one of these crates, and as a consequence, to compile `rasn` dev-dependencies with `arc-slice` feature enabled.


XAMPPRocky:
Thank you for your PR!

PR #489, opened 2025-08-29
Fix field tag stack for XER encoding 

BendixBuchheister:
# Fix `field_tag_stack` of the XER encoder

While integrating the rasn compiler into our project, I came across issues with the XER encoding of some values in special cases.
I found two special cases in which  the `field_tag_stack` of the XER encoder is not behaving as expected.

## CHOICE

When a `CHOICE` is encoded, it didn't remove the current tag from the `field_tag_stack` and thus all elements after the choice were offset by `-1` tag.

### Changes

I have removed the `match` from the `encode_explicit_prefix` function, because the xml_tag was one of the matching arguments, but we shall only pop an element of the `field_tag_stack` in the both Arms which match on `Tag::EOC`.

### Tests

I've added a test `sequence_with_element_after_choice` which shows the behavior. Here is the encoded xml before the changes:

```xml
<SequenceWithChoice>
  <recursion>
    <Leaf />
  </recursion>
  <recursion>
    <wine>
      <true />
    </wine>
    <grappa>00010203</grappa>
    <inner>
      <hidden>
        <false />
      </hidden>
    </inner>
    <oid>1.8270.4.1</oid>
  </recursion>
</SequenceWithChoice>
```

The expected xml would look like this:

```xml
<SequenceWithChoice>
  <recursion>
    <Leaf />
  </recursion>
  <nested>
    <wine>
      <true />
    </wine>
    <grappa>00010203</grappa>
    <inner>
      <hidden>
        <false />
      </hidden>
    </inner>
    <oid>1.8270.4.1</oid>
  </nested>
</SequenceWithChoice>
```

Where the `nested` element of the xml has been named `recursion` instead of `nested`.

## SET OF / SEQUENCE OF

When encoding a `SET OF` or `SEQUENCE OF` each element in the `SET OF / SEQUENCE OF` popped an element from the `field_tag_stack`, which lead to all elements after set `SET OF / SEQUENCE OF` been offset by `+X` tags (with X the number of elements in the `SET OF / SEQUENCE OF`).

### Changes

The `entering_list_item_type` is already set, when we are entering a `SET OF` or `SEQUENCE OF`, so I've made sure to not call `field_tag_stack.pop()` when this flag is set.

### Tests

I've also added two tests `sequence_with_element_after_sequence_of` and `sequence_with_element_after_set_of` which show the behavior. Here is the encoded xml before the changes for `sequence_with_element_after_sequence_of`:

```xml
<SequenceWithSequenceOf>
  <ids>
    <flag>42</flag>
    <int>13</int>
  </ids>
  <enum_val>
    <false />
  </enum_val>
  <int>12</int>
  <enum_val>
    <eins />
  </enum_val>
</SequenceWithSequenceOf>
```

The expected xml would look like this:

```xml
<SequenceWithSetOf>
  <ids>
    <INTEGER>42</INTEGER>
    <INTEGER>13</INTEGER>
  </ids>
  <flag>
    <false />
  </flag>
  <int>12</int>
  <enum_val>
    <eins />
  </enum_val>
</SequenceWithSetOf>
```

Here the elements in the SEQUENCE OF (`ids`) have been named by the tags following in the parent SEQUENCE (`SequenceWithSequenceOf`) after the SEQUENCE OF (so: `flag` and `int`).
Also the names of the elements afterwards are offset, so `flag` is named `enum_val`. The following elements are named correctly again, as the `tag_field_stack` is then completly empty and the identifier passed to the encode function is used as a fallback XML tag.

### DIFF
--- a/src/xer.rs
+++ b/src/xer.rs
@@ -126,6 +126,26 @@ mod tests {
         recursion: Vec<DefaultSequence>,
     }
 
+    #[derive(AsnType, Debug, Encode, Decode, PartialEq)]
+    #[rasn(automatic_tags)]
+    #[rasn(crate_root = "crate")]
+    pub struct SequenceWithSequenceOf {
+        ids: SequenceOfIntegers,
+        flag: bool,
+        int: Integer,
+        enum_val: EnumType,
+    }
+
+    #[derive(AsnType, Debug, Encode, Decode, PartialEq)]
+    #[rasn(automatic_tags)]
+    #[rasn(crate_root = "crate")]
+    pub struct SequenceWithSetOf {
+        ids: SetOfIntegers,
+        flag: bool,
+        int: Integer,
+        enum_val: EnumType,
+    }
+
     fn bool_default() -> bool {
         bool::default()
     }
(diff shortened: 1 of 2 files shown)


6d7a (approved):
LGTM, thank you for your contribution!

PR #493, opened 2025-09-26
fix(per): Fix constrained length encoding/decoding for SIZE(0..65535)

### LINKED ISSUE #492: UPER: Constrained length (0..65535) encoded with short form instead of 16-bit
A SEQUENCE (SIZE(0..65535)) OF T is encoded by using the unconstrained (short-form) length when the span is exactly 65536 (0..65535). According to X.691, when both lower and upper bounds are present and the span (upper − lower + 1) is 65536, the length determinant remains a fixed-width 16‑bit non‑negative binary integer. The encoder was emitting a 1‑octet short form for small lengths in this case. After adjusting the encoder boundary (< SIXTY_FOUR_K → <= SIXTY_FOUR_K), encoding is now correct.

ASN.1 sample used for reproduction
```
World-Schema DEFINITIONS AUTOMATIC TAGS ::=
BEGIN
    ItemList ::= SEQUENCE (SIZE(0..65535)) OF INTEGER
END
```

Test case:
```
#[test]
fn test_minimal_64k_sequence_round_trip() {
    // Round-trip a small sequence to ensure decoder accepts 16-bit constrained length for span 65536
    let original = ItemList(SequenceOf::from(vec![AnonymousItemList(Integer::from(7))]));
    let encoded = rasn::uper::encode(&original).expect("encode");
    // First two bytes should be 0x0001 length determinant (16-bit field)
    assert_eq!(&encoded[0..2], &[0x00, 0x01]);
    let decoded: ItemList = rasn::uper::decode(&encoded).expect("decode");
    assert_eq!(decoded, original);
}
```

The same problem is on decoder side.

dudycz:
Fixes #492 

Length Determinant Rules in PER
According to X.691, for PER encoding of constrained lengths:

Short form: For ranges ≤ 255, uses single octet
16-bit form: For ranges ≤ 65536 (64K), uses 2-octet length determinant
Fragmented form: For ranges > 65536, uses fragmented encoding


XAMPPRocky:
Thank you for your PR! If you rebase the error should be fixed.

XAMPPRocky:
Sorry, that's what I get for not checking, not if you rebase the doc failures should be gone :)

XAMPPRocky:

Also if you can share I am always interested in hearing about what projects you're using rasn in.

PR #497, opened 2025-10-14
BER: don't include full error information when encountering EOF during decoding

### LINKED ISSUE #495: BER decoder performance
I'm currently developing an [EmBER+](https://github.com/Lawo/ember-plus) client and server library in Rust and I'm using rasn for it (EmBER+ uses a ASN.1 specification they call [Glow](https://github.com/babymotte/ember-plus-rs/blob/main/GLOW.asn) and BER encoding rules).

I'm working with a hand-written and rasn-annotated structs since the rasn-compiler was not able to translate the Glow ASN.1 spec, but so far it seems to be correct and I can communicate with devices that speak the EmBER+ protocol.

What I have noticed however is that communication sometimes comes to a complete stop for extended periods of time and I have tracked it down to being caused by the ber decoder taking VERY long to decode larger messages.

Some of the messages I get from devices are as large as >1MB and ber::decode(...) can take up to several minutes to decode one such message on a reasonably new laptop CPU, maxing out one CPU core pretty much the entire time.

The time it takes to decode seems to grow exponentially with the seize of the message, e.g. I have seen a 400kb take~ 20 seconds to decde while a 1.1 MB message took 190 seconds.

I've used cargo flamegraph to try and narrow down the cause further and it seems that almost all of the time is spent debug formatting objects:

Image

Is this a known issue? If not how can I help narrow it down? I'm happy to provide test data.

Nicceboy:
Fixes #495

### DIFF
--- a/src/aper.rs
+++ b/src/aper.rs
@@ -47,7 +47,6 @@ pub fn encode_with_constraints<T: crate::Encode>(
 #[cfg(test)]
 mod tests {
     use crate::{
-        macros::*,
         prelude::*,
         types::{constraints::*, *},
     };
--- a/src/coer.rs
+++ b/src/coer.rs
@@ -79,7 +79,7 @@ pub fn encode_with_constraints<T: crate::Encode>(
 #[allow(clippy::items_after_statements)]
 mod tests {
     use crate::prelude::*;
-    use crate::{self as rasn, constraints, size_constraint, value_constraint};
+    use crate::{self as rasn};
     use bitvec::prelude::*;
     #[test]
     fn bool() {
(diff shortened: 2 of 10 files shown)


Nicceboy:
In this case, EOF was part of the normal decoding logic so the error is not propagated, and it is a bit difficult to test. Decoder could be more explicit with that and it currently may hide real errors. 

> Side note: why is the CI so flaky lately?

There was a bug in nightly compiler, which seems to be fixed now and I fixed the other issues.
 
Do we actually need nightly in pipeline since some changes may never land for stable and it could add extra noise which might not be relevant?

PR #504, opened 2025-10-17
fix(ber): Explicit errors when decoding optional any, proper TLV extraction in sequence

Nicceboy:
- When decoding optional values in BER, does not hide errors anymore and presence status is based on peeking the next tag
- When `Any` type field is in Sequence, the outermost TLV defined by explicit tag is extracted away
- When decoding `Any` outside of Sequence/Set context, TLV pattern is verified but data is just copied as it is. It is up to caller to decode it further with correct type/tag.

### DIFF
--- a/src/ber.rs
+++ b/src/ber.rs
@@ -598,5 +598,36 @@ mod tests {
         ];
         assert_eq!(encode(&value6).unwrap(), expected6);
         assert_eq!(decode::<TestSequence>(expected6).unwrap(), value6);
+
+        // Another type where the final field is optional
+        #[derive(AsnType, Debug, Clone, Encode, Decode, PartialEq)]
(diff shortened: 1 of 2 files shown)


Nicceboy:
I think there is "a bug" that was hidden by the `.ok()` mapping.

In https://github.com/librasn/rasn/blob/6204851c58da76abd85df2a6a39bc65a37925b75/standards/pkix/tests/fuzz.rs#L35-L36 CER attempts to decode following bytes

`30 80 06 01 00 00 00`

as https://github.com/librasn/rasn/blob/6204851c58da76abd85df2a6a39bc65a37925b75/standards/pkix/src/lib.rs#L410-L415 

It does not correctly detect the ending EOC, rather it tried to parse it as tag but failed. It was mapped to `None` previously.

PR #521, opened 2026-01-10
Fix BER encoding of tagged CHOICE

t-higuchi:
This PR fixes #518, #519, and #520.

Fixes for #518 and #519 are trivial.

As for #520, this PR fixes so that `encode()` method of `rasn::Encode` trait is only derived for untagged CHOICE type.
For tagged CHOICE type,  `encode_with_tag_and_constraint()` is derived instead of  `encode()` so that it will handle implicity tagging in the same way where `#[rasn(delegate)]` on `struct Foo(enum)` does.

Decoding is also fixed in the same way.


XAMPPRocky:

PR #522, opened 2026-01-12
Add feature: Implement Default trait for FixedOctetString

ErvinsK:
Hi,

While using the library I discovered that _**FixedOctetString**_ does not implement **Default** trait. Thus, I decided to add a small change to implement **Default** trait for _**FixedOctetString**_ that makes the life a lot of easier since all other **ASN types** have **Default** trait.

//Ervins


XAMPPRocky:

PR #532, opened 2026-01-22
Fix UPER encoding and parsing of extensible CHOICE

jbeyerstedt:
A CHOICE with extensions shall use the number or root variants for the constraints when encoding the index integer. This was already in a comment, but the code below it used the number of all variants instead of the `base_variants` which were already created earlier.

This PR also adds some more test cases for this condition. The existing tests couldn't differentiate the variant index from the padding, so I added some data to each variant.

Fixes #528


XAMPPRocky:
Thank you for your PR! Can you remove the changes to the changelog, it's auto-generated so the formatting will be overwritten or cause issues.

PR #538, opened 2026-02-13
feat(type,jer): implement Encode/Decode on `serde_json::Value`

mkatychev:
Structs like this will faild to derive since `serde_json::Value` (despite being used in the jer module) does not implement Encode/Decode/AsnType:

```rust
#[derive(serde::Serialize, serde::Deserialize, Debug, PartialEq, Eq, Clone, AsnType)]
pub struct Foo {
    inner: serde_json::Value
}
```

Used choice type to implement said traits for `serde_json::Value`.


src/types/json.rs
```diff
@@ -0,0 +1,764 @@
...
+    #[test]
+    fn test_uper() {
+        let value = json!({"name": "test", "value": 42});
+        let encoded = crate::uper::encode(&value).expect("UPER encode failed");
+        let decoded: Value = crate::uper::decode(&encoded).expect("UPER decode failed");
+        assert_eq!(value, decoded);
+    }
+
+    // ============================================================
+    // Test cases adapted from serde-json-canonicalizer https://github.com/evik42/serde-json-canonicalizer
```
mkatychev: part of motivation is that json canonicalization is not available in no_std

XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR!

PR #543, opened 2026-03-11
fix(per): encode empty extension addition group as absent per X.691 §…

Abel981:


XAMPPRocky:
Thank you for your PR!

PR #544, opened 2026-03-12
feat(avn): implement ASN.1 Value Notation codec

jp-marcotte:
Add a new `Codec::Avn` variant implementing the ASN.1 Value Notation text format as defined in ITU-T X.680. AVN is the native textual representation used in ASN.1 modules for DEFAULT values and value assignments, distinct from JER (JSON) and XER (XML).

Architecture mirrors the JER codec:
- `src/avn/value.rs`: `AvnValue` IR enum with `Display` impl for pretty-printed output (indented SEQUENCEs, `'HEX'H`/`'BIN'B` bit strings, `{ arc... }` OIDs, `id : val` CHOICEs)
- `src/avn/enc.rs`: `Encoder` struct preserving SEQUENCE field declaration order via `Vec<(String, AvnValue)>` (vs JER's BTreeMap)
- `src/avn/de.rs`: two-phase decode — recursive-descent lexer/parser producing `AvnValue` tree, then type-directed dispatch; handles space-separated OID arc notation and underscore identifiers
- Error types `AvnDecodeErrorKind` / `AvnEncodeErrorKind` integrated into the existing snafu error hierarchy
- `Codec::Avn` dispatch wired into all five `Codec` methods

Key format rules:
- BOOLEAN: `TRUE` / `FALSE`
- OCTET STRING: `'01FF'H`
- BIT STRING: `'A0'H` (byte-aligned) or `'10'B` (non-byte-aligned)
- OID: `{ 2 5 4 3 }` (space-separated, no commas)
- SEQUENCE: `{\n  field value,\n  ...\n}`
- CHOICE: `identifier : value`
- Absent OPTIONAL fields omitted entirely

Round-trip tests cover: bool, integer, null, octet string, bit string, OID, enumerated, choice, sequence-of, nested sequences with absent optionals, and hyphenated/underscored field identifiers.

### DIFF
--- a/src/avn/de.rs
+++ b/src/avn/de.rs
@@ -1,0 +1,1166 @@
+//! Decoding ASN.1 Value Notation (AVN) text into Rust structures.
+//!
+//! Uses a two-phase approach:
+//! 1. **Lex + parse** the entire input string into an [`AvnValue`] tree.
+//! 2. **Type-directed decode** pops values from a stack and dispatches.
+
+use alloc::collections::BTreeMap;
+use core::str::FromStr;
+use num_bigint::BigInt;
+
+use crate::{
+    Decode,
+    de::Error,
+    error::{AvnDecodeErrorKind, DecodeError},
+    types::{
+        Any, BitString, BmpString, Constraints, Constructed, Date, DecodeChoice, Enumerated,
+        GeneralString, GeneralizedTime, GraphicString, Ia5String, NumericString, ObjectIdentifier,
+        Oid, PrintableString, SetOf, Tag, TeletexString, UtcTime, Utf8String, VisibleString,
+        variants,
+    },
+};
+
+use super::value::AvnValue;
+
+// ---------------------------------------------------------------------------
+// Lexer
+// ---------------------------------------------------------------------------
+
+#[derive(Debug, Clone)]
+enum Token {
+    LBrace,
+    RBrace,
+    Colon,
+    Comma,
+    Identifier(alloc::string::String),
+    Number(alloc::string::String),
+    /// Content of `'...'H` (between the quotes, before the `H`)
+    HexString(alloc::string::String),
+    /// Content of `'...'B` (between the quotes, before the `B`)
+    BinString(alloc::string::String),
+    /// Already-unescaped content of `"..."` string
+    QuotedString(alloc::string::String),
+    True,
+    False,
+    Null,
+}
+
+fn lex(input: &str) -> Result<alloc::vec::Vec<Token>, DecodeError> {
+    let bytes = input.as_bytes();
+    let len = bytes.len();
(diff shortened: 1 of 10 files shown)


src/avn/de.rs
```diff
@@ -0,0 +1,1166 @@
...
+    /// Content of `'...'B` (between the quotes, before the `B`)
+    BinString(alloc::string::String),
+    /// Already-unescaped content of `"..."` string
+    QuotedString(alloc::string::String),
+    True,
+    False,
+    Null,
+}
+
+fn lex(input: &str) -> Result<alloc::vec::Vec<Token>, DecodeError> {
```
XAMPPRocky: Is there a reason not to use `nom` for parsing? We already depend on it for binary parsing, so it wouldn't add any cost to use it here.

src/avn/value.rs
```diff
@@ -0,0 +1,308 @@
...
+    Choice {
+        /// The chosen alternative's identifier
+        identifier: alloc::string::String,
+        /// The value of the chosen alternative
+        value: alloc::boxed::Box<AvnValue>,
+    },
+    /// Any character-string type: `"quoted string"` (X.680 double-quote escaping)
+    CharString(alloc::string::String),
+    /// Sentinel for absent OPTIONAL fields — must not appear in output
+    Absent,
```
XAMPPRocky: For a top level value, this shouldn't be needed right? Shouldn't sequence of be `(String, Option<AvnValue>)`?

XAMPPRocky:
Thank you for your PR! Exciting o see more formats added. What are you using it for if you're comfortable sharing?

jp-marcotte:
> Thank you for your PR! Exciting o see more formats added. What are you using it for if you're comfortable sharing?

AVN is used with Trusted Connectivity Alliance specs, is the industry standard format to describe the content of an eSIM.

src/avn/de.rs
jp-marcotte: no reasons, beside lack of caffeine I guess. Let me change that.

XAMPPRocky:
Interesting! Would you also be interested in contributing some those specs to be added in the `specs` crates directory?

src/avn/de.rs
XAMPPRocky: I mean it will be easier now that you've written it, so it will be easy to rewrite it with the tests ensuring it doesn't regress 🙂

jp-marcotte:
> Interesting! Would you also be interested in contributing some those specs to be added in the `specs` crates directory?

Can you elaborate on this? Do you mean the rasn/standards ?

XAMPPRocky:
Yes sorry, the standards directory. Providing the data types for a standard in a shared crate makes it easier for everyone to contribute and update them

XAMPPRocky:

PR #547, opened 2026-03-15
add ITS standard ETSI TS 102 894-2

schphil:
This PR adds the ITS standard [ETSI TS 102 894-2](https://forge.etsi.org/rep/ITS/asn1/cdd_ts102894_2). 

I was unsure whether it is preferred to include the original ASN.1 sources and generate the Rust code during the build process, or to commit only the generated Rust code. Please let me know if you like this or anything else to be changed.

### DIFF
--- a/standards/its/src/lib.rs
+++ b/standards/its/src/lib.rs
@@ -12,6 +12,7 @@ extern crate alloc;
 #[allow(rustdoc::broken_intra_doc_links)]
 #[allow(rustdoc::invalid_rust_codeblocks)]
 #[allow(rustdoc::bare_urls)]
+#[allow(clippy::doc_overindented_list_items)]
 pub mod ts102894_2;
 
 /// ASN.1 definitions for ETSI TS 103 097
--- a/standards/its/src/lib.rs
+++ b/standards/its/src/lib.rs
@@ -1,6 +1,9 @@
 #![cfg_attr(not(test), no_std)]
 extern crate alloc;
 
+/// ASN.1 definitions for ETSI TS 102 894-2
+pub mod ts102894_2;
+
 /// ASN.1 definitions for ETSI TS 103 097
 pub mod ts103097;
 
--- a/standards/its/src/lib.rs
+++ b/standards/its/src/lib.rs
@@ -1,7 +1,14 @@
 #![cfg_attr(not(test), no_std)]
 extern crate alloc;
 
-/// ASN.1 definitions for ETSI TS 102 894-2
+/// ASN.1 definitions for [ETSI TS 102 894-2].
+///
+/// The standard defines the ITS Common Data Dictionary (CDD), a set of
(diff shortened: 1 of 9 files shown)


standards/its/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 #![cfg_attr(not(test), no_std)]
 extern crate alloc;
 
+/// ASN.1 definitions for ETSI TS 102 894-2
```
XAMPPRocky: Could we include a short description of the standard? Like a couple of sentences explaining what it defines?

schphil: Sure, I’ve added a description.

XAMPPRocky:
For the CI failure, you just need to set it to allow for the module, its generated code so you can't fix it. I would also followup with an issue in the compiler repo, it should escape `[`, `]`.

standards/its/src/lib.rs
```diff
@@ -1,6 +1,20 @@
...
+/// The standard defines the ITS Common Data Dictionary (CDD), a set of
+/// reusable data objects used across ITS application- and facilities-layer
+/// messages. The exact interpretation of these elements is defined in the
+/// corresponding message standards.
+///
+/// [ETSI TS 102 894-2]: https://www.etsi.org/deliver/etsi_ts/102800_102899/10289402/02.01.01_60/ts_10289402v020101p.pdf
+#[allow(rustdoc::broken_intra_doc_links)]
+#[allow(rustdoc::invalid_rust_codeblocks)]
+#[allow(rustdoc::bare_urls)]
+#[allow(clippy::doc_overindented_list_items)]
```
XAMPPRocky:
```suggestion
#[allow(
    rustdoc::broken_intra_doc_links,
    rustdoc::invalid_rust_codeblocks,
    rustdoc::bare_urls,
    clippy::doc_overindented_list_items
)]
#[cfg(not(test))]
```

Nicceboy:
> For the CI failure, you just need to set it to allow for the module, its generated code so you can't fix it. I would also followup with an issue in the compiler repo, it should escape `[`, `]`.

Other ITS modules were partially generated by compiler, but since the output missed so much stuff, they very manually made complete and more usable. Should there be some policy for compiler-only generated code, that it should be only merged if it is fully functional?

XAMPPRocky:
> Should there be some policy for compiler-only generated code, that it should be only merged if it is fully functional?

I mean I would like us to get to the point where the compiler generated code works as expected, and that we are filing bugs against the compiler repo.

I think the only thing I would say around this is keeping the generated code and the handwritten code separate so that the generated code can be regenerated once fixes are merged to the compiler.

schphil:
@XAMPPRocky  do you have any idea why the tests partially failed?

XAMPPRocky:
It looks like formatting errors and documentation not being in comments to me, why only some failed I'm not sure.

PR #548, opened 2026-03-15
WIP: add ITS standard ETSI TS 103301

schphil:
This PR adds the ITS standard [ETSI TS 103 301](https://forge.etsi.org/rep/ITS/asn1/cdd_ts102894_2). 

Note: This currently does not compile.

## Compile errors

1. **Incorrect module imports**

The generated `etsi_its_dsrc` module imports types from `etsi_its_dsrc_region` incorrectly.

Relevant code:

- Import section: https://github.com/schphil/rasn/blob/ts103301/standards/its/src/ts103301/ts103301.rs#L10-L18
- Region type definitions: https://github.com/schphil/rasn/blob/ts103301/standards/its/src/ts103301/ts103301.rs#L6189-L8054

2. **Missing lifetime specifier in `encode`**

The generated encode function lacks the required lifetime bounds for rasn::Encoder, which causes compilation failures with the current encoder trait definition.

Current generator implementation:
https://github.com/librasn/compiler/blob/main/rasn-compiler/src/generator/rasn/builder.rs#L1052-L1062

Proposed fix:
```rust
pub fn encode<E: for<'a> rasn::Encoder<'a>>(&self, encoder: &mut E, identifier: & #class_unique_id_type_name) -> Result<(), <E as rasn::Encoder<'_>>::Error> {
    match (self, identifier) {
        #(#en_match_arms)*
        _ => Err(rasn::error::EncodeError::from_kind(
            rasn::error::EncodeErrorKind::Custom {
                msg: alloc::format!("Unknown unique identifier for information object class instance."),
            },
            encoder.codec()
        ).into())
    }
}
```

3. **Inconsistent `RegionId` type generation**

In some generated regional extension structs, the correct alias `RegionId` is used, while in others the type is flattened to the primitive u8. This leads to compile errors because `decode`expects `RegionId` type. 

Example:
https://github.com/schphil/rasn/blob/ts103301/standards/its/src/ts103301/ts103301.rs#L301-L325

## Build output
[build_output.txt](https://github.com/user-attachments/files/26005969/build_output.txt)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,12 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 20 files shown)


schphil:
@XAMPPRocky Do you have any quick guesses/notes on the errors before I look into them in more detail? Many thanks in advance.

PR #549, opened 2026-03-19
feat: support named_values for INTEGER NamedNumberList in derive macros and AVN codec

jp-marcotte:
## Summary
  - Add `named_values` support to `Constraints` so the AVN encoder outputs symbolic identifiers (e.g. `pukAppl1`) instead of bare decimals for INTEGER types
   with a NamedNumberList, per ITU-T X.680 §19.9
  - Parse `#[rasn(named_values("name" = val, ...))]` in the AsnType derive macro so the rasn-compiler can emit this attribute and have it flow through to
  runtime constraints
  - AVN decoder accepts both named identifiers and bare integer values
  - JER and DER codecs are unaffected

  ## Test plan
  - [ ] AVN encode: INTEGER value with known name emits the identifier (e.g. `1` → `pukAppl1`)
  - [ ] AVN encode: INTEGER value without a name emits bare decimal
  - [ ] AVN decode: accepts named identifier and bare integer
  - [ ] Derive macro round-trip: `#[rasn(named_values(...))]` is parsed and produces correct `Constraints`
  - [ ] JER/DER codecs unchanged

### DIFF
--- a/macros/macros_impl/src/config.rs
+++ b/macros/macros_impl/src/config.rs
@@ -41,9 +41,10 @@ impl Constraints {
             return None;
         }
         // Emit: named_values("pukAppl1" = 1, "pukAppl2" = 2, ...)
-        let pairs = self.named_values.iter().map(|(val, name)| {
-            quote!(#name = #val)
-        });
+        let pairs = self
+            .named_values
+            .iter()
+            .map(|(val, name)| quote!(#name = #val));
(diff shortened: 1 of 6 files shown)


XAMPPRocky:
Thank you for your PR! What is the current status of this? It was marked as ready to review, but there's a test plan in the description, so it's not clear if it's complete.

PR #553, opened 2026-04-13
COER: Fix validation of Integer with length determinant encoding

mickdermack:
The current version of the validation logic for Integer values with length determinant encoding in COER has at least two bugs:
* The unsigned number 0 is considered invalid
* Signed Integer values are not validated to be canonical

This PR fixes these bugs and adds a few more test vectors to avoid regressions.

### DIFF
--- a/src/coer.rs
+++ b/src/coer.rs
@@ -231,6 +231,7 @@ mod tests {
             (i128::from(u64::MAX) + 1).try_into().unwrap(),
             &[0x09, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]
         );
+        round_trip!(coer, E, E::new(0), &[0x01, 0x00]);
         round_trip!(coer, F, F::new(2), &[0x00, 0x02]);
         // Error expected, outside of range constraints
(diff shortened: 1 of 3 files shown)


XAMPPRocky (approved):
Thank you for your PR! Looks good to me, just needs to be formatted.

mickdermack:
Actually, I think I made a mistake in the signed case... I guess I shouldn't try to do ASN.1 while sleep-deprived.

EDIT: The check seems to be OK, just the tests are broken.
EDIT2: I think the library contains yet another bug, which hid the broken test case. Let's see if I can fix this one too...

mickdermack:
I added another commit addressing the following bug:
The sizing of OER encoded Integers was not always correct. For example, an Integer with the bounds [i64::MIN - 1, i64::MAX + 1] would incorrectly be encoded as 64 bits signed, even though neither the start nor the end value fit in an i64.

Nicceboy:
Looks good to me! Missed these when doing the original implementation.

Nicceboy (approved):

XAMPPRocky:

PR #557, opened 2026-04-23
fix(pkix): omit revoked_certificates in empty CRLs

KoffeinFlummi:
First off, thank you very much for your work with this library! Longtime user, firsttime contributor.

According to RFC 5280 section 5.1.2.6:

> When there are no revoked certificates, the revoked certificates list
> MUST be absent.

I'm not entirely confident `rasn(default)` is the proper way to handle this, but it works.


XAMPPRocky:

The CI failures are unrelated to your changes.

Also if you're using Rasn and feel comfortable sharing I always love hearing what people are using it in? 🙂

PR #559, opened 2026-04-27
fix(xer): Return XER error on integer overflow when decoding

jbeyerstedt:
Just a small copy-paste typo I stumbled over.

XER integer decoding shall give an error with `Codec::Xer` instead of `Codec::Jer` (like it's done in the other decoding methods)


XAMPPRocky:
Thank you for your PR!

PR #560, opened 2026-04-29
feat: Add feature flags to enable specific codecs (to optimize application size)

jbeyerstedt:
The size of an ASN.1 parser library can get quite big when `rasn` is generating code for all supported codecs. This can be an issue on embedded targets where XER and JER encoding may not be needed when you're only operating on the rust types.

This PR adds a set of feature flags to individually enable each codec "suite":
- `codec_ber`: BER, CER and PER
- `codec_oer`: OER and COER
- `codec_per`: UPER and APER
- `codec_xer`: XER
- `codec_jer`: JER
- `codec_avn`: AVN

Sadly, all codes depend on some BER and DER encoding/ decoding which means that full BER/ CER/ DER support is always carried even if e.g. only UPER is needed. But it wasn't easily achievable to only pull in the required BER and DER parts.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,12 +28,20 @@ chrono = { version = "0.4.38", default-features = false, features = ["alloc"] }
 bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 
 [features]
-default = ["f32", "f64", "bytes"]
+default = ["f32", "f64", "bytes", "codecs"]
 std = []
 f32 = []
 f64 = []
 backtraces = ["std", "snafu/backtrace"]
 compiler = ["rasn-compiler"]
+# shorthand to enable all supported codecs at once
+codecs = ["codec_ber", "codec_oer", "codec_per", "codec_jer", "codec_xer", "codec_avn"]
+codec_ber = []
+codec_oer = ["codec_ber"]
+codec_per = ["codec_ber"]
+codec_jer = ["codec_ber", "dep:serde_json"]
+codec_xer = ["codec_ber"]
+codec_avn = ["codec_ber"]
 
 [profile.bench-lto]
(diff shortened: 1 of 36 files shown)


XAMPPRocky:
Thank you for your PR! While I'm generally positive on this idea. I would want some validation on your claim that the current structure affects binary size. It's not disputable that it affects compile times because there are dependencies which are only used by certain codecs, but I'm not sure how it would be affecting binary size since any code that isn't used should be eliminated.

Additionally I would just name the feature flags `per`, `jer`, `xer`, etc. Having the `codec_` prefix is not needed in my opinion, if you're using the library you understand that you are enabling a codec.

jbeyerstedt:
> but I'm not sure how it would be affecting binary size since any code that isn't used should be eliminated.

That's what I also thought, but at least with standard settings of an ESP-HAL project it did affect the artifact size. But for some reason they also build with "debug = 2" in the release profile (and no "strip" setting), so I just experimented a bit more with the "debug" and "strip" parameters using my ESP32 project:

- before this PR:
  - `strip = "none"`, `debug = 2`, : 29 MB
  - `strip = "none"`, `debug = false`: 3.3 MB
  - `strip = true` (`debug` doesn't matter): 1.8 MB
- with this PR:
  - `strip = "none"`, `debug = 2`, : 16 MB
  - `strip = "none"`, `debug = false`: 2.6 MB
  - `strip = true` (`debug` doesn't matter): 1.2 MB

So the main difference is made by stripping all symbols (which is kinda obvious and I already stripped debug symbols before). But in the end, it still makes a difference somehow.
(I must admit though, that "removing" the XER and JER encoding also removes two little pieces of string search and replace in the `c-its-parser` library, but I don't think that this makes a difference of 0.6MB.)

PR #562, opened 2026-07-15
fix(der): reject trailing data in der::decode

### LINKED ISSUE #552: This can be an attack surface: der::decode() silently accepts trailing bytes — signature bypass vector
Hi rasn maintainers,

I'm reporting a (potential) security issue in rasn's `der::decode()` that affects all versions through 0.28.11.
(seems I cannot get a email addr, and now there is not private issues, so I can only report it here)

## Summary
`rasn::der::decode()` does not verify that all input bytes are consumed after decoding. X.690 DER §8.1.1.1 mandates that a DER encoding is exactly one complete TLV with no trailing data. Currently, `der::decode()` delegates to `T::decode()` and returns the result without checking `decoder.remaining()`:
```rust
// src/der.rs:6-11
pub fn decode<T: crate::Decode>(input: &[u8]) -> Result<T, crate::error::DecodeError> {
    T::decode(&mut crate::ber::de::Decoder::new(
        input,
        crate::ber::de::DecoderOptions::der(),
    ))
    // ← No check that all input was consumed
}
```
This means `rasn::der::decode::<Null>(&[0x05, 0x00, 0xDE, 0xAD, 0xBE, 0xEF])` succeeds silently, accepting 4 trailing garbage bytes.

## Security Impact
This bug is structurally identical to several assigned CVEs for the same class of issue:
| CVE | Lib | CVSS | Bug |
| ----- | ---- | -------- | ----    |
|CVE-2018-16150	 | axTLS | High | Trailing bytes in PKCS#1 → Bleichenbacher forgery |
|CVE-2018-16151/16152 | strongswan | High | Trailing bytes in ASN.1 → signature bypass |
|CVE-2014-1569	| NSS | Critical | DER permissiveness enables data smuggling |

Concrete attack scenarios:
1. Bleichenbacher-style signature forgery: If rasn is used to parse decrypted PKCS#1 v1.5 DigestInfo, trailing bytes after the ASN.1 structure are silently accepted. With RSA e=3, an attacker can forge signatures.
2. Two-parser chain bypass: If a strict DER parser signs/verifies bytes and rasn parses the same input, appended data is invisible to rasn but changes the byte representation — enabling signature verification bypass.
3. Round-trip divergence: decode(input) then encode(value) produces different bytes than the original, breaking certificate transparency logs, replay detection, and any system comparing raw DER encodings.

## Reproduction
```rust
use rasn::der;
use rasn::types::Null;

let input = &[0x05, 0x00, 0xDE, 0xAD, 0xBE, 0xEF]; // NULL + trailing garbage
let result = der::decode::<Null>(input);
assert!(result.is_ok()); // BUG: should reject trailing bytes
```

## Discovery
Found via source code audit during differential fuzzing 

I saw you provide a interface for [rasn-pkix](https://github.com/librasn/rasn/tree/67dd50f2e18afa42c847bf3b614745c07af5f6c0/standards/pkix), so I think this should be reported as a security issue.
Looking forward to your reply.

ivmat:
## Summary

Fixes #552.

`der::decode` decoded the root value and returned without checking that the whole input
was consumed, so `der::decode::<()>(&[0x05, 0x00, 0xDE, 0xAD, 0xBE, 0xEF])` (a `NULL`
followed by four garbage bytes) decoded to `()` and silently dropped the trailing bytes.
X.690 §8.1.1.1 requires a DER message to be exactly one complete value with no trailing
data. The same leniency is what produced the round-trip divergence @tynus2 flagged, where
`encode(decode(bytes))` can be shorter than `bytes`. I've kept out of the
security-classification discussion and scoped this purely to the exact-consumption / API
contract of `decode<T>`.

## What the fix does

`src/der.rs`: after decoding the root value, `der::decode` now checks the decoder's
remaining input and returns `DecodeErrorKind::UnexpectedExtraData { length }` when any
bytes are left. It reuses the extra-data error rasn already emits for trailing bytes inside
a constructed value (`src/ber/de.rs`), so the behaviour and error type stay consistent with
the rest of the codec. `decode_with_remainder` is unchanged, for callers that intentionally
decode a value from the front of a larger buffer, and the `decode` doc comment now states
the strictness.

## Scope: DER only, and the codec-wide question

I scoped this to `der::decode`, matching the issue. @Nicceboy's point that the same pattern
applies to several `decode` entry points is right: each codec has its own thin top-level
`decode` wrapper (`ber`, `cer`, `oer`, `uper`, ...), and the strictness lives in that
wrapper rather than in shared machinery, so extending it is mechanical. I deliberately did
not touch the others here:

- BER has historically dropped trailing bytes (as you noted, established practice given its
  streaming heritage), so making it strict is a semantics decision I didn't want to take
  unilaterally.
- CER is a distinguished form like DER and is a natural next candidate; OER/PER are the
  same shape of question.

If you'd like the same check applied uniformly across the strict `decode` entry points
(with the `decode` docs updated to state the strictness, per your suggestion), I'm happy to
do that as a follow-up — just say which codecs you want it on.

## Real-world impact surfaced by the change

Two fixtures already in the tree carried trailing bytes that the lenient decoder silently
accepted, and both are legitimate:

- `standards/cms/tests/data/pesig.p7` (Authenticode): the PKCS#7 blob lives inside a
  WIN_CERTIFICATE structure padded to an 8-byte boundary, so it has two `00` padding bytes
  after the DER value. `test_authenticode` now parses it with `decode_with_remainder`.
- `standards/pkix/tests/data/splice.bin` (a fuzz-regression fixture): a valid
  `AlgorithmIdentifier` followed by three trailing bytes. The `splice` test's DER arm now
  uses `decode_with_remainder`, matching its existing BER/CER arms.

Both are small, mechanical migrations to the "decode one value, ignore the rest" API, and
are included here so CI stays green. They're also a useful signal for the codec-wide
decision: trailing bytes do turn up in real inputs (container padding), so the strict
default and the `decode_with_remainder` escape hatch matter together.

## Tests

`tests/issue552.rs`:

- `der_decode_rejects_trailing_bytes` — the issue's exact `NULL` + 4 trailing bytes now
  returns `UnexpectedExtraData { length: 4 }`.
- `der_decode_rejects_trailing_bytes_on_integer` — `INTEGER` + 1 trailing byte.
- `der_decode_accepts_exact_input` — exact encodings still decode.
- `der_decode_round_trip_holds_for_exact_input` — `encode(decode(bytes)) == bytes` for
  exact input.

I confirmed the two rejection tests fail on the unpatched tree and pass with the fix. The
full workspace suite is green (`cargo test --workspace`), as are `cargo fmt --all --
--check`, `cargo clippy --workspace --all-targets --features=f32,f64,bytes,std,backtraces
-- -D warnings`, and the doc build with `-D warnings`.

## Evidence note (independent reference)

As an independent cross-check on the property, not as a dependency: I maintain
`der-verified` (0.1.0, MIT/Apache-2.0, proofs re-runnable from a fresh clone), a DER/X.690
core in which exact whole-input consumption is one of the verified properties. Its
top-level strict decoders return a distinct `TrailingData` error on any trailing bytes, and
a Kani harness proves that whenever the TLV reader accepts an input it consumes exactly
`header + declared_length` bytes and never over-reads. That is bounded model checking (a
16-byte symbolic buffer, loop unwinding to depth 16) rather than an all-length statement,
but it covers the trailing-data property on that domain; the crate also carries round-trip
harnesses of the same shape. This PR simply brings `der::decode` in line with that
exact-consumption behaviour.

### DIFF
--- a/src/der.rs
+++ b/src/der.rs
@@ -3,11 +3,25 @@
 pub use crate::ber::*;
 
 /// Attempts to decode `T` from `input` using DER.
+///
+/// `input` must contain exactly one complete DER value; any trailing bytes left
+/// after the top-level value are rejected, as required by X.690 §8.1.1.1. Use
+/// [`decode_with_remainder`] to decode a value from the front of a larger buffer.
+///
+/// # Errors
+/// Returns `DecodeError` if `input` is not valid DER encoding specific to the
+/// expected type, or if there is unexpected trailing data after the decoded value.
 pub fn decode<T: crate::Decode>(input: &[u8]) -> Result<T, crate::error::DecodeError> {
-    T::decode(&mut crate::ber::de::Decoder::new(
-        input,
-        crate::ber::de::DecoderOptions::der(),
-    ))
+    let decoder = &mut de::Decoder::new(input, de::DecoderOptions::der());
+    let decoded = T::decode(decoder)?;
+    let remaining = decoder.remaining();
+    if !remaining.is_empty() {
+        return Err(crate::error::DecodeError::unexpected_extra_data(
+            remaining.len(),
+            decoder.codec(),
+        ));
+    }
+    Ok(decoded)
 }
 /// Attempts to decode `T` from `input` using DER. Returns both `T` and reference to the remainder of the input.
 ///
--- a/standards/cms/tests/test_cms.rs
+++ b/standards/cms/tests/test_cms.rs
@@ -1,5 +1,5 @@
 use pretty_assertions::assert_eq;
-use rasn::der::{decode, encode};
+use rasn::der::{decode, decode_with_remainder, encode};
 
 use rasn_cms::authenticode::{
     SPC_CLASS_UUID, SPC_INDIRECT_DATA_OBJID, SpcIndirectDataContent, SpcLink, SpcPeImageData,
@@ -39,7 +39,12 @@ fn test_cms_encrypted() {
 
 #[test]
 fn test_authenticode() {
-    let info = decode::<ContentInfo>(PE_SIG_DATA).unwrap();
+    // Authenticode stores the PKCS#7 blob in a WIN_CERTIFICATE structure padded
+    // to an 8-byte boundary, so `PE_SIG_DATA` carries two trailing padding bytes
+    // after the DER value. `der::decode` is now strict and rejects trailing bytes
+    // (see #552), so use `decode_with_remainder` to parse the complete ContentInfo
+    // and ignore that container padding.
+    let (info, _padding) = decode_with_remainder::<ContentInfo>(PE_SIG_DATA).unwrap();
     assert_eq!(CONTENT_SIGNED_DATA, info.content_type);
 
     let signed_data = decode::<pkcs7_compat::SignedData>(info.content.as_bytes()).unwrap();
--- a/standards/pkix/tests/fuzz.rs
+++ b/standards/pkix/tests/fuzz.rs
@@ -35,7 +35,12 @@ fn splice() {
     let encoded = rasn::cer::encode(&value).unwrap();
     assert_eq!(value, rasn::cer::decode(&encoded).unwrap());
 
-    let value = rasn::der::decode::<rasn_pkix::AlgorithmIdentifier>(data).unwrap();
+    // `splice.bin` carries trailing bytes after the AlgorithmIdentifier. DER
+    // decoding is now strict and rejects trailing data (see #552), so parse the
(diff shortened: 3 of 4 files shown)


ivmat:
The matrix failures are pre-existing: current stable clippy's new `byte_char_slices` lint fires in `src/types/strings/{numeric,printable}.rs`, which this PR doesn't touch — main's last green CI run (May 4) predates the lint. Split the fix out into #563.

XAMPPRocky:
should be fixed by rebasing now :)

PR #564, opened 2026-07-23
Fix APER byte-alignment for constrained strings inside CHOICE + SEQUENCE(OPTIONAL) nesting

CharlesShan-hub:
### Problem

APER encoding produces corrupted data (PermittedAlphabetError on decode) when a CHOICE containing a constrained `VisibleString` is nested inside a SEQUENCE with OPTIONAL fields. The same issue also affects any non-extensible SEQUENCE (with OPTIONAL fields) nested inside a CHOICE.

### Root Cause

1. **`encode_choice` byte-alignment offset** — When computing `preceding_bits` for the choice encoder's `parent_output_length`, the count included `number_optional_default_fields` but ignored `preamble_pre_reserved`. In the fast path for non-extensible sequences (introduced in edition 2024), preamble bits are pre-allocated in `self.output`, so adding `number_optional_default_fields` double-counts them. This causes `pad_to_alignment` to compute the wrong byte boundary, shifting constrained string data by a few bits and producing invalid visible string characters on decode.

2. **Fast path drops `parent_output_length`** — When a non-extensible SEQUENCE is encoded via the fast path (buffer reuse optimization), the child encoder was created with `parent_output_length: None`, discarding the parent's preceding-bit info. For SEQUENCE types nested inside a CHOICE variant, this means byte-alignment calculations inside the SEQUENCE (e.g., for constrained strings) fail to account for the CHOICE index bits and outer OPTIONAL bitmap, leading to the same class of decode errors.

### Fix

1. In `encode_choice`: subtract `preamble_pre_reserved` from the `preceding_bits` calculation, matching the pattern already used in `output_length()`.

2. In the non-extensible SEQUENCE fast path: propagate `self.parent_output_length` to the child encoder instead of setting it to `None`.

### Testing

- `cargo test --lib` (rasn): 331 passed, 0 failed
- Full Java test suite (DL/T 2811 CMS protocol via JNI): 319 passed, 0 failures, 0 errors

### DIFF
--- a/src/per/de.rs
+++ b/src/per/de.rs
@@ -288,9 +288,6 @@ impl<'input, const RFC: usize, const EFC: usize> Decoder<'input, RFC, EFC> {
             if range == 0 {
                 Ok(input)
             } else if range == 1 {
-                if self.options.aligned {
-                    input = self.parse_padding(input)?;
-                }
                 (decode_fn)(input, size_constraint.minimum())
             } else {
                 let range = if self.options.aligned && range > 256 {
@@ -761,6 +758,17 @@ impl<'input, const RFC: usize, const EFC: usize> crate::Decoder for Decoder<'inp
         let mut octet_string = Vec::new();
         let codec = self.codec();
 
+        // Aligned PER (X.691 §17): fixed-size OCTET STRING with SIZE > 2 is
+        // octet-aligned. The encoder calls pad_to_alignment before the length
+        // determinant, so we must consume that padding here too.
+        if let Some(size) = constraints.size() {
+            if size.constraint.range() == Some(1) && size.constraint.as_start() > Some(&2) {
+                if self.options.aligned {
+                    self.input = self.parse_padding(self.input)?;
+                }
+            }
+        }
+
         self.decode_extensible_container(constraints, |input, length| {
             let (input, part) = nom::bytes::streaming::take(length * 8)(input)
                 .map_err(|e| DecodeError::map_nom_err(e, codec))?;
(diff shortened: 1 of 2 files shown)


XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR! seems it just needs ti be formatted :)

CharlesShan-hub (commented):
thank you！

XAMPPRocky:
```
failures:

---- extensible_aper stdout ----

thread 'extensible_aper' (8586) panicked at tests/personnel.rs:459:1:
called `Result::unwrap()` on an `Err` value: DecodeError { kind: PermittedAlphabetError { reason: IndexNotFound { index: 14 } }, codec: Aper, backtrace: Backtrace(()) }


failures:
    extensible_aper

test result: FAILED. 13 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

XAMPPRocky:
@CharlesShan-hub friendly ping :)

CharlesShan-hub:
Thanks for your patience and for the review! Here's what changed since your last look:

1. Fixed the extensible_aper failure — the range == 1 branch in decode_string_length must consume octet-alignment padding before reading a fixed-size (SIZE(n)) constrained string. It had been accidentally removed alongside the OCTET STRING alignment fix, so I restored it.
2. Formatted the code with cargo fmt.
3. Synced with the latest upstream/main so the branch is up to date with librasn/main (v0.28.14).
4. 
All tests pass locally, including extensible_aper. Ready for re-review.

XAMPPRocky:

PR #567, opened 2026-08-07
fix(per): validate decoded value and size constraints

### LINKED ISSUE #501: No Length/Integer Range check for UPER and APER decoders
Hi, 

I found that UPER and APER's decoder does not enforce the size constraint. In contrast, the encoder will enforce those constraints. 

For example (in the UPER's case), given this ASN.1 message

`S1 ::= INTEGER (1..10)`

If we give an the generated parser input `\xf0`, it will accept the input and parse it to 16. But we cannot encode this value and will output 

`Error: EncodeError { kind: ValueConstraintNotSatisfied { value: 16, expected: Range { start: Some(1), end: Some(10) } }, codec: Uper }`

Similar for Sequence-Of, 

`S2 ::= SEQUENCE (SIZE (1..10)) OF BOOLEAN`

A input like `\xff\xff\xff\xff\xff\xff` will be parsed to a boolean sequence of length 16, but we cannot encode the message with error:
`Error: EncodeError { kind: InvalidLength { length: 16, expected: Range { start: Some(1), end: Some(10) } }, codec: Uper }`

I have not tested other range constraints like OCTET STRING or BIT STRING but I suspect similar behaviors. 

To improve, I think the decoder should enforce these constraint checks to be consistent with the encoders.

dudycz:
## Summary

- reject UPER and APER integer values decoded from surplus constrained code points
- validate decoded string and container lengths against their root size constraints
- preserve extensible constraints by skipping root validation when the extension bit is set
- cover the issue examples plus aligned, half-open, string, and extensible constraint cases

Closes #501

### DIFF
--- a/src/per/de.rs
+++ b/src/per/de.rs
@@ -179,11 +179,15 @@ impl<'input, const RFC: usize, const EFC: usize> Decoder<'input, RFC, EFC> {
         mut decode_fn: impl FnMut(InputSlice<'input>, usize) -> Result<InputSlice<'input>>,
     ) -> Result<()> {
         let extensible_is_present = self.parse_extensible_bit(constraints)?;
-        let constraints = constraints.size().filter(|_| !extensible_is_present);
+        let size = constraints.size().filter(|_| !extensible_is_present);
+        let mut total_length = 0usize;
         let input =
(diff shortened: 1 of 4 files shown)


XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR!

Issue #488, opened 2025-08-22
Any ignores context tag


Huntragon:
I was implementing the OSI Association Control Service (more specifically, the AARQ) and I'm only interested in a part of the sequence, so I created the entire struct and used `Option<Any>` for all of the fields that are optional but I'm not currently interested in (which might change in the future).
When doing this, I noticed that `Option<Any>` always decodes correctly, even though the context tag does not match.
I've created the next reproducible example:
```rs
use rasn::prelude::*;

#[derive(Debug, AsnType, Decode, Encode, PartialEq, Eq)]
struct IncomingRequest {
    #[rasn(tag(Context, 0))]
    handshake: Option<Any>,
    #[rasn(tag(Context, 1))]
    user_data: Option<Any>,
}

fn main() {
    let incoming: IncomingRequest = rasn::Codec::Ber
        .decode_from_binary(&[0x30, 0x06, 0xA1, 0x04, 0x01, 0x02, 0x03, 0x04])
        .unwrap();
    let expected = IncomingRequest {
        handshake: None,
        user_data: Some(Any::new(vec![0x01, 0x02, 0x03, 0x04])),
    };
    assert_eq!(incoming, expected);
}
```
The assertion fails because the library decodes the `handshake` part of the incoming request, even though the context tag that is in the payload is `1`.
Furthermore (this might be a design decision, I'm not sure), the inner contents of the `Any` type that is decode contain the tag, length and value, instead of discarding the tag and length and keeping the value as an opaque sequence of bytes.

If this is intended behaviour, I'm willing to create a MR and add an `Opaque` (or something similar) type that will check the tag correctly and keep the opaque bytes.

XAMPPRocky:
Thank you for your issue! I don't think that is intended behaviour, I would also expect that the tag checked.

XAMPPRocky:
Breaking the type signature is an acceptable fix. We are still pre-1.0 for this reason.

XAMPPRocky:
What if `Any` also carried its identifier so it allows inspection and then the identifier validation can be done after decoding in the decode impl? That would be codec agnostic.

PR #569, opened 2026-08-11
fix BER encoding of tagged ANY type

t-higuchi:
rasn 0.28.14 fails round-trip test for BER encoding of tagged ANY type since it does not correctly encode the constructed bit in the tag.

This bug breaks the use case outlined in #488, which involves replacing data types that is not interested in with `ANY`.

This PR adds a round-trip tests to reproduce the issue and one-line fix for it. 
Existing tests in src/ber.rs are also fixed since they did not meet the ITU-T X.680 standard.

### DIFF
--- a/src/ber.rs
+++ b/src/ber.rs
@@ -573,9 +573,11 @@ mod tests {
             d: true,
         };
         // Sequence tag (0x30), length 7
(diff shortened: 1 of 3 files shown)


XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR!

